PR #4, opened 2021-09-02
Dev/agp/fix warnings

Gnistrup:
Some warnings was quite bad and could cause some undefined behaviour.

### DIFF
--- a/STM32/DC/InputValidation/Src/inputValidation.c
+++ b/STM32/DC/InputValidation/Src/inputValidation.c
@@ -37,7 +37,7 @@ int getNumberDigits(int inputNumber){
 bool isInputInt(char * inputBuffer, int startIdx, int endIdx) {
 	int idx = startIdx;
 	while (idx <= endIdx) {
-		if (!isdigit(inputBuffer[idx])) {
+		if (!isdigit((int)inputBuffer[idx])) {
 			if (inputBuffer[idx] == ' ' && idx != startIdx){
 				break;
 			}
@@ -92,7 +92,7 @@ struct actuationInfo parseAndValidateInput(char * inputBuffer){
 	}
 
 	// Check if X in pX is int and valid
-	if (!isdigit(inputBuffer[1]) || atoi(&inputBuffer[1]) > 4 || atoi(&inputBuffer[1]) == 0){
+	if (!isdigit((int)inputBuffer[1]) || atoi(&inputBuffer[1]) > 4 || atoi(&inputBuffer[1]) == 0){
 		return info;
 	} else {
 		info.pin = atoi(&inputBuffer[1]) - 1; // User uses 1-indexed ports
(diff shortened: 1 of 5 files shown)


STM32/DC/InputValidation/Src/inputValidation.c
```diff
@@ -37,7 +37,7 @@ int getNumberDigits(int inputNumber){
 bool isInputInt(char * inputBuffer, int startIdx, int endIdx) {
 	int idx = startIdx;
 	while (idx <= endIdx) {
-		if (!isdigit(inputBuffer[idx])) {
+		if (!isdigit((int)inputBuffer[idx])) {
```
matias-cphatomics, line 40: Will this not cause a casting error if it is not a digit? I think it states in the documentation of isdigit() that you have to pass a char.

STM32/DC/InputValidation/Src/inputValidation.c
```diff
@@ -92,7 +92,7 @@ struct actuationInfo parseAndValidateInput(char * inputBuffer){
 	}
 
 	// Check if X in pX is int and valid
-	if (!isdigit(inputBuffer[1]) || atoi(&inputBuffer[1]) > 4 || atoi(&inputBuffer[1]) == 0){
+	if (!isdigit((int)inputBuffer[1]) || atoi(&inputBuffer[1]) > 4 || atoi(&inputBuffer[1]) == 0){
```
matias-cphatomics, line 95: Same as above.

STM32/DC/USB_DEVICE/App/usbd_cdc_if.c
```diff
@@ -247,19 +247,13 @@ static int8_t CDC_Control_FS(uint8_t cmd, uint8_t* pbuf, uint16_t length)
...
 
-    case CDC_SET_CONTROL_LINE_STATE:{
+    case CDC_SET_CONTROL_LINE_STATE: {
         USBD_SetupReqTypedef * req = (USBD_SetupReqTypedef *)pbuf;
-        if(req->wValue &0x0001 != 0){
-      	  isComPortOpen = true;
-        } else {
-          isComPortOpen = false;
-        }
+        isComPortOpen = ((req->wValue & 0x0001) != 0);
```
matias-cphatomics, line 252: Are we sure the only  states  (req->wValue is & 0x0001) can take are 0 and 1? If yes, then fine. If no, then we should check this as the isComPortOpen could be set wrongly.

STM32/DC/InputValidation/Src/inputValidation.c
matias-cphatomics: I read the documentation wrong, my bad :)

matias-cphatomics (approved):

STM32/DC/USB_DEVICE/App/usbd_cdc_if.c
Gnistrup: Yep I am sure. (req->wValue & 0x0001) is logic bit operator, i.e. only the last bit will remain.

PR #7, opened 2021-09-13
Dev/agp/fix ac board

Gnistrup:
Moved AC board to use dual DMA buffer handling.
Also made USBprinf obsolete.


matias-cphatomics:
When saying "made USBprintf obsolete" I'm guessing you simply mean for the public repository and have not removed any functionality since private repo is still dependent on it.

Otherwise it looks fine :)

PR #9, opened 2021-09-14
Use circualar buffer when sending via USB cdc

Gnistrup:
This will make the HAL delay obsoletewhen sending via USB cdc


Gnistrup:
I have just found out that the complete USB_DEVICE is duplicated in ALL projects.
This is far from optimal since every change must be made in all project.

matias-cphatomics:
Has this PR become obsolete after #20 which is more general? :)

Gnistrup:
Fixed in commit f731ad7bc694c961560df0c6f605e923434f6fd7.
This commit was only solving this issue pr. project, f731ad7bc694c961560df0c6f605e923434f6fd7 made it common.

PR #12, opened 2021-09-17
Deleted launch files and added to gitignore

Gnistrup:
Also added .gitattributes, *.ioc, *.cproject, *.mxproject is binary.
This will reduce the git diff to source files.


matias-cphatomics:
It seems like the changes in pull-request #9 has been mixed in with this one. I agree with the entries added to the gitignore, but would like to approve the USB CDC changes in the other pull request.

Gnistrup:
Changed only to include the remove off files.

matias-cphatomics (approved):

PR #13, opened 2021-09-21
Temperature board: General clean up

matias-cphatomics:
Adding GPIO ports/pins to arrays and accessing via loops. Improves readability and cuts ~100 lines of code.

### DIFF
--- a/STM32/Temperature/Core/Src/main.c
+++ b/STM32/Temperature/Core/Src/main.c
@@ -283,11 +283,14 @@ int main(void)
     /* USER CODE BEGIN 3 */
 
 	// Upload data every "tsUpload" ms.
-	if (HAL_GetTick() - timeStamp > tsUpload){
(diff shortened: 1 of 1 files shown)


STM32/Temperature/Core/Src/main.c
```diff
@@ -355,21 +283,19 @@ int main(void)
     /* USER CODE BEGIN 3 */
 
 	// Upload data every "tsUpload" ms.
-	if (HAL_GetTick() - timeStamp > tsUpload)
-	{
+	if (HAL_GetTick() - timeStamp > tsUpload){
```
Gnistrup, line 286: I am not a big fan of having '{' on the same line as the if statement. The scope of the 'if' will be a bit more clear if moved to the next (code is more readable). But, we just have to agree on a style.

matias-cphatomics: I personally prefer the other, but since you like the other better I am happy to switch :)

Issue #14, opened 2021-09-23
Add overcurrent shutoff on AC and DC boards. 


tomgossland-cphatomics:
The AC board was not designed to exceed continuous 8A current through each port. The code should monitor this and shutoff the port if this is exceeded. It may also be a good idea to monitor heatsink temperature and disable or reduce switching frequency if heatsink exceeds 50C. 

The DC board was not designed to exceed 2A continuous current through each port including the motor driver port (A destructive test needs to be done before increasing this limit). The code should monitor and enforce this as well. The board temperature should also limit or disable switching if it exceeds 50C.

These cases should send a message reporting that this safety feature is being implemented.

PR #19, opened 2021-09-27
AC/DC board: Max current draw and max temp detection and handling

matias-cphatomics:
The AC board is not designed to withstand continuous current draw above 9A. For the DC it is 2A. If a current draw on either of the boards exceed those limits the boards will turn off all pins and give a warning. Warning format:

"warning: max current draw exceeded."

Likewise for temperature the boards are not designed to withstand high temperature for longer periods of time. If the board temperature exceeds 50C then all pins are shut off and a warning is given.

"warning: max temp exceeded."

### DIFF
--- a/STM32/DC/Core/Src/main.c
+++ b/STM32/DC/Core/Src/main.c
@@ -63,6 +63,9 @@
 #define TURNONPWM 999
 #define TURNOFFPWM 0
 
+#define MAX_CURRENT_DRAW 2
+#define MAX_BOARD_TEMP 50
+
 //-------------F4xx UID--------------------
 #define ID1 (*(unsigned long *)0x1FFF7A10)
 #define ID2 (*(unsigned long *)0x1FFF7A14)
@@ -150,10 +153,25 @@ void printHeader()
     USBnprintf(buf);
 }
 
+void shutChannelOffAndReport(uint16_t channel){
+	turnOffPin(channel);
+	USBnprintf("warning: max current exceeded.");
+}
+
+void shutAllOffAndReport(){
+	allOff();
+	USBnprintf("warning: max temp exceeded.");
+}
 
 static double meanCurrent(const int16_t *pData, uint16_t channel)
 {
-    return current_scalar * ADCMean(pData, channel) + current_bias;
+    double meanCurrent = current_scalar * ADCMean(pData, channel) + current_bias;
+
+    // If component draws more current than max allowed then operation is shut off
+    if (meanCurrent > MAX_CURRENT_DRAW){
+    	shutChannelOffAndReport(channel);
+    }
+    return meanCurrent;
 }
 
 void printResult(int16_t *pBuffer, int noOfChannels, int noOfSamples)
@@ -167,6 +185,11 @@ void printResult(int16_t *pBuffer, int noOfChannels, int noOfSamples)
         clearLineAndBuffer();
 
     const double temp = si7051Temp(&hi2c1);
+    // If board temp exceeds 50C shut down operation
+    if (temp>MAX_BOARD_TEMP){
+    	shutAllOffAndReport();
+    }
+
     USBnprintf("%.2f, %.2f, %.2f, %.2f, %.2f, %.2f, %d",
             meanCurrent(pBuffer, 0), meanCurrent(pBuffer, 1), meanCurrent(pBuffer, 2), meanCurrent(pBuffer, 3),
             temp, ADCrms(pBuffer, 5), ADCmax(pBuffer, 5));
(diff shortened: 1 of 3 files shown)


matias-cphatomics:
Fixes #14 but warnings should also be handled in LoopControl @freddyrios

freddyrios:
`The AC board is not designed to withstand continuous current draw above 9A. For the DC it is 2A. If a current draw on either of the boards exceed those limits the boards will turn off all pins and give a warning`

@matias-cphatomics is there a reason not to keep the other ports operational when one of the ports shorts or is given too big of a load? Also how long does the feature survives, until next comms restart? until power cycling the board? if not, what's the procedure after replacing the problem, for example a valve.

`Likewise for temperature the boards are not designed to withstand high temperature for longer periods of time. If the board temperature exceeds 50C then all pins are shut off and a warning is given.`

when does it become operational again? when going below a certain treshold, say 45C? 
extra thought: of course I agree we should not let boards run wild outside the operating temperature range, but on the long run we (cph atomics) and anybody using the boards for other purposes need to be extremely aware of this in any safety context. That is, scenarios that can cause 50C should be evaluated to ensure the lost of all actuators hooked to the ports do not make a bad situation worse (which happens regardless of the changes because doing the actuation has its own risks).

matias-cphatomics:
@freddyrios I agree this is not developed enough. Some of the following actions mentioned could well be handled in the LoopControl, but of course we would have to look at that simultaneously if that is the case.

1. The ports could be shut of individually with very little change. 
2. Right now they simply shut of until the next actuation so that a surrounding control system can take action as a reaction on the warnings. It is therefore only a temporary shut off. This is the case both for the current draw and temperature. 
3. Right now it only shuts off pins, but does not put them on again to their previous state after returning to a normal state. This could be done with temperature measurements, but would not make sense for the current case.

Let's talk together about how we can implement this in cooperation so that it can function both as a stand-alone application and also with LoopControl.

STM32/DC/Core/Src/main.c
```diff
@@ -151,9 +154,25 @@ void printHeader()
...
+}
+
 static double meanCurrent(const int16_t *pData, uint16_t channel)
 {
-    return current_scalar * ADCMean(pData, channel) + current_bias;
+    double meanCurrent = current_scalar * ADCMean(pData, channel) + current_bias;
+
+    // If component draws more current than max allowed then operation is shut off
+    if (meanCurrent > MAX_CURRENT_DRAW){
+    	bool isCurrentExceeded = true;
```
Gnistrup, line 172: I am not a big fan of putting this shut off logic in the simple functions meanCurrent etc.
When looking at the name of function it's indicate that it performs a simple calculation. Now it is also handling system logic.

Suggest to save values in printValues. This will also make it possible to list which valve it drawing to much current/temp.

freddyrios:
> 1. The ports could be shut of individually with very little change.
> 2. Right now they simply shut of until the next actuation so that a surrounding control system can take action as a reaction on the warnings. It is therefore only a temporary shut off. This is the case both for the current draw and temperature.
> 3. Right now it only shuts off pins, but does not put them on again to their previous state after returning to a normal state. This could be done with temperature measurements, but would not make sense for the current case.

@matias-cphatomics it seems doing point 1 and keeping point 3 unchanged make sense. For the later it does make sense to require explicit action to resume operation after an incident. Anyway a control system and/or operator should be made aware of something wrong is going on, so it seems reasonable. If timing of recovery is important then it can be automated by a control system and potentially an alternative solution must exist anyway (depending on safety and operational requirements).

point 2 vs. explicit recovery sounds like a trade off, so maybe we should keep it as is for now. Assuming trying to resume when the problem remains would retrigger and shut down again, then the difference is control systems that don't handle the warning/recovery would stay off (if they don't have retries or keep on commands) or keep sending some reattempts that would retrigger the issue vs. the board staying off.

The below remains but can't see any way around it. That is, the boards will have operating conditions, so the overall system design must account for this. These limits must be made visible, both internally and to any external consumer of the boards. 

> extra thought: of course I agree we should not let boards run wild outside the operating temperature range, but on the long run we (cph atomics) and anybody using the boards for other purposes need to be extremely aware of this in any safety context. That is, scenarios that can cause 50C should be evaluated to ensure the lost of all actuators hooked to the ports do not make a bad situation worse (which happens regardless of the changes because doing the actuation has its own risks).

tomgossland-cphatomics:
This warning could also be used by the loopcontrol to switch to a second valve if we've added one for redundancy.

freddyrios:
> This warning could also be used by the loopcontrol to switch to a second valve if we've added one for redundancy.

Depends on the purpose and approach of the redundancy. Too big of a topic, let's take that one separately.

PR #20, opened 2021-09-27
Use common circualar buffer when sending via USB cdc

Gnistrup:
This will make the HAL delay obsolete when sending upper/lower layer
design pattern. The new usb_cdc_fops module is also common for all modules.

### DIFF
--- a/STM32/Libraries/USBprint/Src/USBprint.c
+++ b/STM32/Libraries/USBprint/Src/USBprint.c
@@ -10,7 +10,11 @@
 #include "stdarg.h"
 #include "string.h"
 #include "stdio.h"
-#include "usbd_cdc_if.h"
+#ifdef USE_COMMON_USB_CDC
+    #include "usb_cdc_fops.h"
+#else
+    #include "usbd_cdc_if.h"
+#endif
 
 int USBnprintf(const char * format, ... )
 {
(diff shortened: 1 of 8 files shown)


STM32/Libraries/USBprint/Src/usb_cdc_fops.c
```diff
@@ -0,0 +1,275 @@
...
+  * @retval no of bytes received.
+  */
+void usb_cdc_rx(uint8_t* buf)
+{
+    *buf = 0;
+    if (usb_cdc_if.rx.ctx)
+        circular_read_command(usb_cdc_if.rx.ctx, buf);
+}
+
+void usb_cdc_flush()
```
matias-cphatomics, line 191: Consider renaming to something like "usb_rx_flush" since it only concerns receiver side or use an input argument to allow clearing of transmit side - even though I'm not sure when it would be relevant to flush transmit buffer :)

STM32/Libraries/USBprint/Src/USBprint.c
```diff
@@ -22,8 +26,12 @@ int USBnprintf(const char * format, ... )
     len += vsnprintf(&buffer[len], sizeof(buffer) - len, format, args);
     va_end (args);
 
+#ifdef USE_COMMON_USB_CDC
```
matias-cphatomics, line 29: Good temporary work around :) Looking forward to seeing only "include usb_cdc_fops.h" in there soon.

matias-cphatomics (approved):
Good job! Looking forward to a demo.

PR #33, opened 2021-11-10
Dev/agp/issue14

Gnistrup:
simplified code handling GPIO's


STM32/AC/HeatCtrl/Src/HeatCtrl.c
```diff
@@ -0,0 +1,148 @@
...
+        // Use modules '%' to get the on/off section in PWM period.
+        // If percent is 0, heat shall be off (period is invalid)
+        if (pCtrl->pwmPercent == 0)
+        {
+            pCtrl->heater->set(pCtrl->heater, false);
+        }
+        else
+        {
+            uint32_t periodOn = (pCtrl->pwmPercent * pCtrl->pwmPeriod) / 100;
+            tdiff = tdiff % pCtrl->pwmPeriod;
```
lt-copenhagenatomics, line 82: A comment would be nice here :-)

STM32/AC/HeatCtrl/Src/HeatCtrl.c
```diff
@@ -0,0 +1,148 @@
...
+static uint32_t tDiff(uint32_t head, uint32_t tail)
+{
+    if (head >= tail)
+        return head - tail;
+
+    // Overflow of head:
+    // (head + 2^32) - tail = (2^32 - tail) + head = (see below)
+    return (((uint32_t) -1) - tail) + 1 + head;
+}
+
```
lt-copenhagenatomics, line 52: If I remember correct then this method is the only one that changes heater output. This is "critical" to your design. I would add that in a comment.

lt-copenhagenatomics (changes requested):

PR #34, opened 2021-11-18
Issue 62, Flash Calibration sector must be reserved in ld linker script.

Gnistrup:
This should be done to ensure that write of calibration data will not
clear sections of the program data (.text etc).


Gnistrup:
Testet på SaltFlow da DC/AC og Temperatur ikke bruger skriv af calibration data.

lt-copenhagenatomics:
Only comment it that code could be moved to last 128 k area.

PR #38, opened 2021-12-30
Added initial version of the OTP programmer

Gnistrup:
This version can not do much since it is a skeloton. It can receive commands via
USB and reply Serial and jumt to bootloader. Next version require changes in CA Protocol
and a new utility able to read from OTP area in device.


Gnistrup:
Basicly, it is only the otpApp that is new. The rest is just skeleton code generated by STM ide.

PR #48, opened 2022-03-09
Issue #75, No need to make avarage on temperature board.

Gnistrup:
Not needed since the aquiring time in the ADC is quite slow and works OK.
Further, if an additional filtering should be made, this can be done in the connected host system.

### DIFF
--- a/STM32/Libraries/SPI/Src/ADS1120.c
+++ b/STM32/Libraries/SPI/Src/ADS1120.c
@@ -283,7 +283,6 @@ void ADS1120Loop(ADS1120Device *dev)
     else
     {
         // The smallest possible State machine
-        double temp;
         switch(data->currentInput)
         {
         case INPUT_TEMP:
@@ -291,24 +290,16 @@ void ADS1120Loop(ADS1120Device *dev)
             break;
 
         case INPUT_CHA:
-            temp = adc2Temp(((int16_t) adcValue), data->calibration);
-            if (temp == 10000 || data->chA == 10000)
-                data->chA = temp;
-            else
-                data->chA += (temp - data->chA)/mAvgTime;
+            data->chA = adc2Temp(((int16_t) adcValue), data->calibration);
             break;
 
         case INPUT_CHB:
-            temp = adc2Temp(((int16_t) adcValue), data->calibration);
-            if (temp == 10000 || data->chB == 10000)
-                data->chB = temp;
-            else
-                data->chB += (temp - data->chB)/mAvgTime;
+            data->chB = adc2Temp(((int16_t) adcValue), data->calibration);
             break;
 
         case INPUT_CALIBREATE:
             // The offset value from ADC1120 can be either negative or positive.
-            data->calibration = (((int16_t) adcValue) - data->calibration) / mAvgTime;
+            data->calibration += (((int16_t) adcValue) - data->calibration) / mAvgTime;
             break;
         }
         setInput(dev, nextInput(data->currentInput), false);


freddyrios:
> Not needed since the aquiring time in the ADC is quite slow and works OK. Further, if an additional filtering should be made, this can be done in the connected host system.

does this comment means we are sampling every 50 ms and is that what you are referring to? // current setting is lowest sample rate, 1/20 Sec. Timeout is 70mSec.

Just a bit confused about the context of this. By the way, there still seems to be usages of moving averages in some of the other lines. The reason I find it relevant, is because the code links to this https://en.wikipedia.org/wiki/Moving_average, but I am uncertain about which of the specific approaches its using and whether its correct i.e.
- simple moving average: drops the oldest point (we are clearly not doing that)
- cumulative moving average: the denominator n+1 is an increasing number as one adds values to the cumulative moving average

Gnistrup:
The 1/20 is the acquiring time used in the ADC and since this is a SPI device there is a timeout.
The reading of ADC in SPI devices runs using a event based method since all 5 SPI devices must run in parallel and when done write the result to an internal memory location. This commit removes the moving average method.
The report via USB done using a time calculation using the values saved in memory.

STM32/Libraries/SPI/Src/ADS1120.c
```diff
@@ -283,32 +283,23 @@ void ADS1120Loop(ADS1120Device *dev)
...
-                data->chB = temp;
-            else
-                data->chB += (temp - data->chB)/mAvgTime;
+            data->chB = adc2Temp(((int16_t) adcValue), data->calibration);
             break;
 
         case INPUT_CALIBREATE:
             // The offset value from ADC1120 can be either negative or positive.
-            data->calibration = (((int16_t) adcValue) - data->calibration) / mAvgTime;
+            data->calibration += (((int16_t) adcValue) - data->calibration) / mAvgTime;
```
freddyrios, line 302: suggest to triple check the use of a moving average, because the current approach does not seem to match neither simple moving average nor cumulative moving average. Same for the internalTemp earlier.

freddyrios (approved):
aproved, but still think the remaining moving averages need attention

STM32/Libraries/SPI/Src/ADS1120.c
Gnistrup: It does, taken from https://en.wikipedia.org/wiki/Moving_average
CMA(n+1) = CMA(n) + (X(n+1) - CMA(n))/(n+1)
equeals
CMA(n+1) += (X(n+1) - CMA(n))/(n+1)
Which is implemented as:
data->calibration += (((int16_t) adcValue) - data->calibration) / mAvgTime
same for temperature.

freddyrios: the way I read that is that:
CMA(2) = CMA(1) + ((X(2) - CMA(1)) / **1**
CMA(3) = CMA(2) + ((X(3) - CMA(2)) / **2**
CMA(10) = CMA(9) + ((X(10) - CMA(10)) / **10**

That is, the n+1 represents how many points have already been gathered so far. Also note the example given " For example, an investor may want the average price of all of the stock transactions for a particular stock up until the current time", it means the cumulative average applies to *all* data, while it sounds like the intention was for it to only apply to the last 6 data points.

One alternative is to stick to the simple moving average and keep a circular buffer of the last 6 data points like they suggest, so that on each cycle the oldest value gets removed from the moving average.

Gnistrup: ??? That is not possible since CMA(10) is not yet calculated on the right side.

freddyrios: mispelled, meant: CMA(10) = CMA(9) + ((X(10) - CMA(9)) / **10**

freddyrios: so the existing code is instead doing: CMA(10) = CMA(9) + ((X(10) - CMA(9)) / **6**

Gnistrup: That is correct and is how it should be. The 6 is called the window, in this case 5+1, N=5.
But lets talk tomorrow.

freddyrios: @Gnistrup please take a look at this https://docs.google.com/spreadsheets/d/1RI_cEEEjVskZdXI3f8sdlLTgB1gYploSvLYVK_hAM7o/edit?usp=sharing

it shows how different that value is to the simple moving average (which it feels its what its trying to approximate). Pay special attention to the end of the table.

freddyrios: @Gnistrup wait, had a mistake, just a moment

freddyrios: @Gnistrup fixed the table. 

With stable data the change is not very relevant, but with changing data it has a lot of resistance to change. Its not too bad, as I guess the new data having a weight of 1/6th does make a sizable contribution. 

Still take a look, you can see I made a big change at row 60 and then look at the deviation column. It takes over 15 samples to get reasonably close to the new value and 40 samples to get below a 0.01 difference.

freddyrios: outcome of offline discussion *and* TL:DR; of whole thread
- we are using specifically this one [Modified moving average / running moving average / smoothed moving average](https://en.wikipedia.org/wiki/Moving_average#Modified_moving_average) with N = 6, which a version of an [Exponential moving average](https://en.wikipedia.org/wiki/Moving_average#Modified_moving_average#Exponential_moving_average)
- N = 6 seems to have a slower reaction compared to a single moving average of 6, even early on the first 2 samples after the change. This is positive for calibration as big changes are most likely to be due to electrical noise i.e. value should be slow changing + it contributes very little to the final temperatures. The internal temperature are used to get the overall box temperature, and these are also expected to be slow changing.
- the board reads temperatures at 1/20 and these are already averaged underneath. It 2 reads x chip (as there are 2 temperatures attached to each), so we only have 1 sample x 10 Hz cycle (as anders said, these filters then can happen at the PI)

PR #53, opened 2022-03-30
Dev/agp/issue54

Gnistrup:
Moved Port GPIO control to CAProtocol


STM32/Libraries/UnitTesting/src/stubs.cpp
```diff
@@ -102,34 +107,63 @@ int calCompare(int noOfPorts, const CACalibration* catAr)
...
+        return calCompare(noOfPorts, calAray);
+    }
+
+private:
+    CAProtocolCtx caProto;;
+    static std::queue<uint8_t> testString;
+    static int testReader(uint8_t* rxBuf)
+    {
+        *rxBuf = testString.front();
+        testString.pop();
```
lasse-tarp-CA, line 175: In line 127 after push there is a input++. Should there be a input-- here?

Gnistrup: input is a pointer. It is pointer into a string  (array of chars) that is pushed into a queue.

matias-cphatomics:
Tested on both AC and DC boards. Works as expected. Fixes issue 54 in CA_embedded_private.

PR #55, opened 2022-04-06
Update SHT35 library

matias-cphatomics:
Removal of static functions definition in header file. 

Until now it has only caused warnings, but is may potentially cause some errors.


matias-cphatomics:
Closing as this only has effect on projects using the SHT35 humidity chip. Only one project uses them, which can not build until these changes takes affect.

PR #57, opened 2022-04-08
Generic temperature reading + User specification of sensor type.

matias-cphatomics:
All temperature sensor readings are now done according to the NIST lookup tables. At this point only type K and J are supported.

The board takes an input command of 10 characters each of which specifies a sensor i.e. J or K at this point.
Ex.

JJJJJKKKKK

Tells the board that the first 5 inputs are Type J thermocouples and the last 5 are Type K. 

Additional sensor types can be added with minimal effort. Simply find the Delta and Cold-Junction Delta for the specific type and add to the temperature.h file. Then associate that type with a corresponding letter :)


STM32/Temperature/Core/Src/Temperature.c
```diff
@@ -84,6 +92,7 @@ void InitTemperature(SPI_HandleTypeDef* hspi_)
     if (getPcbVersion(&ver) || ver.major != 5 || ver.minor < 2)
```
lasse-tarp-CA, line 92: This will not work for version 6+

PR #69, opened 2022-08-12
LightController: Initial project

matias-cphatomics:
The light controller has three ports that controls RGB led lights.

**EDIT:** It takes the following input:  portNumber, 6 character HEX number i.e. p2 AAFF00 would set port 2 to have red: AA (170), green: FF (256) and blue 00 (0).
 
It prints the 3 channels with 10Hz frequency comma separated i.e.:

HexP1, HexP2, HexP3 -> e.g. FF0000, AAFF00, BBCC11

**Files of interest: main.c/h, LightController.c/h**


freddyrios:
I assume these colors are about the same as rgb colors in css: https://www.w3schools.com/colors/default.asp

freddyrios (approved):
what's the plan when using it with LoopControl?

I am mainly thinking that the simple route of uploading the colors as they come as separate values would take 3 x 8 bytes (as things work today), when in reality the full range of colors is just 3 bytes.

matias-cphatomics:
> what's the plan when using it with LoopControl?
> 
> I am mainly thinking that the simple route of uploading the colors as they come as separate values would take 3 x 8 bytes (as things work today), when in reality the full range of colors is just 3 bytes.

You are right, the communication is a bit wasteful. I can rewrite it to take three connected hex values if you prefer. :+1:

PR #70, opened 2022-08-16
Library: Flash read/write using CRC32 for data integrity 

matias-cphatomics:
The FLASH read/write library has been updated to include 'safe' versions of the data storing/retrieving.

The #ifdef clauses ensure that the built-in CRC module of the STM32F401 MCU has been enabled, since a project will not have access to the CRC handle, HAL_CRC_Calculate function as well as the stm32f4xx_hal_crc.h file otherwise. 

Note: The data stored in using the built in FLASH functions take uint8_t data type as input, whereas the CRC32 module only accepts uint32_t. Hence, memcpy calls are necessary since a simple conversion between uint8_t and uint32_t violates the [strict aliasing rule](https://stackoverflow.com/questions/98650/what-is-the-strict-aliasing-rule) i.e. please pay extra attention to those calls to ensure the code can not go into any unwanted/unexpected behavior. 

The boards will be updated afterwards to include the CRC module and the legacy functions may be removed after all boards using FLASH functionality has been updated and call the update 'safe' functions.

### DIFF
--- a/STM32/Libraries/FLASH_readwrite/Inc/FLASH_readwrite.h
+++ b/STM32/Libraries/FLASH_readwrite/Inc/FLASH_readwrite.h
@@ -8,9 +8,16 @@ Description:		Provides an interface to read and write FLASH memory.
 #define INC_FLASH_READWRITE_H_
 
 #include <stdint.h>
+#ifdef HAL_CRC_MODULE_ENABLED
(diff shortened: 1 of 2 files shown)


lasse-tarp-CA (commented):
Guess that danger is shared memory. Don't have C skills that could find any problems here

PR #73, opened 2022-09-30
AC: Overheat prevention.

matias-cphatomics:
An overheat prevention sequence is implemented. 

If the temperature reaches the MAX_TEMPERATURE then the board reduces its current PWM duty cycle until the board has reached the  MAX_TEMPERATURE again.

During the time when the board has exceeded the  MAX_TEMPERATURE the board disregards any input from a user that tries to increase the PWM duty cycle of the pins. If the user sends a command below the current PWM duty cycle i.e. to decrease temperature or shut off when the board has exceeded the  MAX_TEMPERATURE the command is accepted as normally.

### DIFF
--- a/STM32/AC/Core/Src/ACBoard.c
+++ b/STM32/AC/Core/Src/ACBoard.c
@@ -213,8 +213,10 @@ static void heatSinkLoop()
     {
     	unsigned long now = HAL_GetTick();
     	if (now - previous > 2000)	// 2000ms. Should be aligned with the control period of heater.
+    	{
+    		previous = now;
     		adjustPWMDown();
-    	previous = now;
+    	}
     }
 }
 
--- a/STM32/AC/Core/Src/ACBoard.c
+++ b/STM32/AC/Core/Src/ACBoard.c
@@ -22,6 +22,8 @@
 #define ADC_CHANNELS	5
 #define ADC_CHANNEL_BUF_SIZE	400
 
+#define MAX_TEMPERATURE 80
+
 /* GPIO settings. */
 static struct
 {
@@ -182,11 +184,19 @@ static void CAallOn(bool isOn)
 }
 static void CAportState(int port, bool state, int percent, int duration)
 {
+	uint8_t pwmPercent;
+	getPWMPinPercent(port-1, &pwmPercent);
+	// If heat sink has reached the maximum allowed temperature and user
+	// tries to heat the system further up then disregard the input command
+	if (heatSinkTemperature > MAX_TEMPERATURE && percent > pwmPercent)
+		return;
+
     actuatePins((ActuationInfo) { port - 1, percent, 1000*duration, true });
 }
 
 static void heatSinkLoop()
 {
+	static unsigned long previous = 0;
     // Turn on fan if temp > 55 and turn of when temp < 50.
     if (stmGetGpio(fanCtrl) && isFanAutoOn)
     {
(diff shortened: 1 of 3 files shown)


STM32/AC/Core/Src/ACBoard.c
```diff
@@ -182,22 +184,39 @@ static void CAallOn(bool isOn)
...
         if (heatSinkTemperature < 50)
         {
             stmSetGpio(fanCtrl, false);
         }
-        else if (heatSinkTemperature > 55)
-        {
-            stmSetGpio(fanCtrl, true);
-        }
+    }
+    else if (heatSinkTemperature > 55)
```
freddyrios, line 208: should this also be checking isFanAutoOn?

also I suggest to get rid of the nested if above, although that might require explicitely checking the getgpio result in both if

STM32/AC/HeatCtrl/Src/HeatCtrl.c
```diff
@@ -128,3 +128,24 @@ void setPWMPin(int pin, int pwmPct, int duration_ms)
...
 }
+
+void adjustPWMDown()
+{
+    for(HeatCtrl *ctx = heaters; ctx < &heaters[noOfHeaters]; ctx++)
+    {
+    	if (ctx->pwmPercent >= 1)
+    	{
+    		ctx->pwmPercent -= 1;
+    		ctx->pwmDuration = MAX_DURATION;
```
freddyrios, line 139: does this value has any relation to the on duration the LoopControl sends?

STM32/AC/HeatCtrl/Src/HeatCtrl.c
```diff
@@ -128,3 +128,24 @@ void setPWMPin(int pin, int pwmPct, int duration_ms)
...
+    	}
+    }
+}
+
+void getPWMPinPercent(int pin, uint8_t *pwmPercent)
```
freddyrios, line 144: why not returning the pwmPercent as the method's return value? i.e. just wondering if there is an advantage of passing an address around vs. the value

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -182,22 +184,39 @@ static void CAallOn(bool isOn)
 }
 static void CAportState(int port, bool state, int percent, int duration)
 {
+	uint8_t pwmPercent;
+	getPWMPinPercent(port-1, &pwmPercent);
+	// If heat sink has reached the maximum allowed temperature and user
+	// tries to heat the system further up then disregard the input command
+	if (heatSinkTemperature > MAX_TEMPERATURE && percent > pwmPercent)
```
freddyrios, line 191: how fast is the temperature of the switchboard reacting to the adjustments it makes?

asking because if its very slow, it sounds like here we would accept the off command of the LoopControl's heater control cycle and then reject the on commands until the switchboard is at the right temperature. So if its slow, the pwm adjust down is only used in the first LoopControl's heater cycle and then it remains off until it gets to the switchboard temperature.

I guess that *if* LoopControl's have not been set a max power *and* the switchboard oveheat happens far from the oven target then the pwm adjustments is likely to rule most of the time. Otherwise the behavior I mentioned above is likely to happen, but only if the switchboard temperature reaction is very slow. 

ps. remember for anyone changing from 20-30% to 100% power to carefully review their physical's system sizing (power cable, fuses) + their pcontrol values in IO.conf (they could see bigger overshoot as max power plays a part on slowing down the rising temperature).

freddyrios (approved):
*if* you want the engaging of this feature to gain some visibility, you could print a very low frequency message when it first triggers (ideally max. once x 10 mins).

STM32/AC/HeatCtrl/Src/HeatCtrl.c
matias-cphatomics: No advantage really, but I did realise some problems arising from doing the shift i.e. when the user specifies port 0 (meaning targeting all ports) the method will not work. Will make a fix.

STM32/AC/HeatCtrl/Src/HeatCtrl.c
matias-cphatomics: Yes and no. 'pwmDuration' is the amount of ms the pwm as a whole should run. The switching on/off comes from pwmPercent and pwmPeriod. (line 63 of HeatCtrl.c). When setting pwmDuration to max it keeps running the pwm signal indefinitely with the same frequency as set from LoopControl. 

If we do not set pwmDuration it will shut off after the last period is sent by LoopControl and therefore not stay on the output that gives the maxTemperature.

PR #106, opened 2023-07-04
Test: Code generation test

matias-cphatomics:
A new static code test is added for the STM32 projects. 

It checks that the DMA is initialised before the ADC component. If the ADC is initialised first it will never start. This wrong initialisation is caused by a bug in the STMCubeMX  auto-generation of code.

It is called whenever a change to an STM32 project is made.

### DIFF
--- a/.github/workflows/stm32build.yml
+++ b/.github/workflows/stm32build.yml
@@ -71,18 +71,18 @@ jobs:
         fi
   run_static_tests:
     needs: changes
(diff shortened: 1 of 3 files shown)


.github/workflows/stm32build.yml
```diff
@@ -69,6 +69,20 @@ jobs:
           echo "Test failed"
           exit 1
         fi
+  run_static_tests:
```
freddyrios, line 72: An alternative is to run the static tests as an extra step of run_tests, so the already running environments with the checked out code is used.

freddyrios (approved):

PR #108, opened 2023-08-31
Feature/aircon update

LukeWalker-CA:
Updated aircon transmitter module to send messages in both old and new AC format.

* Refactored transmitterIR.c file to functionalise sending messages (should make it easier to just send one code if we later decide that is a better course of action)
* Added codes for new AC unit
* Made it so the old messages are sent first then the new messages sent second automatically


STM32/AirconCtrl/Core/Inc/transmitterIR.h
```diff
@@ -40,7 +50,43 @@
 #define FAN_HIGH	0xa0000000
 #define FAN_HIGH_5	0xa000a800
 
-// Sending PWMs
+#define MSG_LEN_BITS 112
+
+/* IR codes AC2 */
+#define AC2_TEMP_COMMAND 0xB24D
+
+#define AC2_FAN_MODE_0 0x9F60
```
matias-cphatomics, line 58: It looks like AC2_FAN_MODE_0/1/A are not being used. If that's the case I think we should remove those entries.

LukeWalker-CA: I think its useful to keep these "register defines" since they don't consume any space after compilation

PR #117, opened 2023-09-29
AC board Status

LukeWalker-CA:
* Added over-temperature error and port status bits

The AC board now outputs a 32-bit status code (hex) at the end of each data transmission every 10Hz. The 16 MSB bit are defined by the status library (part of systemInfo.c/.h). The AC board specific status code is located in the LSB and are defined as follows:   

|5 (MSB) | 4 | 3 | 2 | 1 | 0 (LSB) |
| --- | --- | --- | --- | --- | --- |
| Version Error | Port 4 On/Off | Port 3 On/Off | Port 2 On/Off | Port 1 On/Off | Fan On/Off |

The status bits display the actual on/off status of the port or fan. E.g. during PWM they will flicker on/off depending on the PWM duration.

The "Status" command displays verbose On/Off status + the current PWM level.

* Fixed a bug where the safety mode would never kick in
@matias-cphatomics Should version error be in the library repo perhaps?


STM32/AC/Core/Src/ACBoard.c
```diff
@@ -188,63 +224,111 @@ static void CAallOn(bool isOn)
...
+        unsigned long now = HAL_GetTick();
+        if (now - previous > 2000)  // 2000ms. Should be aligned with the control period of heater.
+        {
+            previous = now;
+            adjustPWMDown();
+        }
+
+        bsSetError(BS_OVER_TEMPERATURE_Msk);
     }
-    else if (heatSinkTemperature > MAX_TEMPERATURE) // Safety mode to avoid overheating of the board
```
LukeWalker-CA, line 212: I noticed a bug here: 
Example, if heatSinkTemperature = 75 we want to start reducing PWM.
```
        if (heatSinkTemperature < 50 && !isFanForceOn) // false, doesn't trigger
        {
            stmSetGpio(fanCtrl, false);
        }
        else if (heatSinkTemperature > 55) // true, triggers
        {
            stmSetGpio(fanCtrl, true);
        }
        else if (heatSinkTemperature > MAX_TEMPERATURE) // Never gets here
        {
            // Reduce PWM
        }
```
The bug is corrected now, but this will introduce "new" behaviour

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -188,63 +224,113 @@ static void CAallOn(bool isOn)
...
 {
-	// Always allow for DFU also if programmed on non-matching board or PCB version.
-	initCAProtocol(&caProto, usb_cdc_rx);
+    // Always allow for DFU also if programmed on non-matching board or PCB version.
+    initCAProtocol(&caProto, usb_cdc_rx);
 
     BoardType board;
     if (getBoardInfo(&board, NULL) || board != AC_Board)
+    {
+        bsSetError(AC_BOARD_VERSION_ERROR);
```
matias-cphatomics, line 310: Going into this condition (and not executing the rest of the init function) means that the ADC is never initialised. Hence, the ADCMonitorLoop never triggers the printCurrentArray callback function and the status is therefore never printed.

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -188,63 +224,113 @@ static void CAallOn(bool isOn)
...
         return;
+    }
 
     // Pin out has changed from PCB V6.0 - older versions need other software.
     pcbVersion ver;
     if (getPcbVersion(&ver) || ver.major < 6)
+    {
+        bsSetError(AC_BOARD_VERSION_ERROR);
```
matias-cphatomics, line 318: Same as the previous comment.

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -188,63 +224,113 @@ static void CAallOn(bool isOn)
...
+    setBoardTemp(heatSinkTemperature);
+}
+
+/*!
+** @brief Updates the global error/status object.
+**
+** Adds the port and fan status bits into the error field, and clears the error bit if there are no
+** more errors.
+*/
+static void updateErrorField()
```
matias-cphatomics, line 282: Maybe a better name would be "updateBoardStatus" since it also relates to states that are not error related.

STM32/AC/Core/Src/ACBoard.h
```diff
@@ -8,7 +8,28 @@
...
-void handleUserInputs(const char* startMsg);
+/***************************************************************************************************
+** DEFINES
+***************************************************************************************************/
+
+/* AC board status register definitions */
+
+/* Each of the 5 ports (1x Fan + 4x AC) can be on or off */
+#define AC_BOARD_PORT_x_STATUS_Msk(x) (1U << (x))
+#define AC_BOARD_PORTS_STATUS_Msk     ((1U << 5U) - 1U)
```
matias-cphatomics, line 19: Is this Msk used?

matias-cphatomics:
I like the idea that all of the states can be handled within a single function. I'm thinking whether it makes sense to set all of the port x on/off status bits in the functions that sets/resets those ports. 

Pros: This would ensure that there is no delay between setting/resetting the ports and having set the states. 
Cons: Perhaps more cluttered code.  

Note: I'm unsure whether it makes a noticeable difference as we only output data with 10Hz and the program runs considerably faster than that so for the vast majority of the cases the output would be similar. There could be one print cycle difference.

STM32/AC/Core/Src/ACBoard.c
LukeWalker-CA: OK, sorted

STM32/AC/Core/Src/ACBoard.h
LukeWalker-CA: Nope, but it feels right to me. I'll delete if you care strongly.

LukeWalker-CA:
> I like the idea that all of the states can be handled within a single function. I'm thinking whether it makes sense to set all of the port x on/off status bits in the functions that sets/resets those ports.
> 
> Pros: This would ensure that there is no delay between setting/resetting the ports and having set the states. Cons: Perhaps more cluttered code.
> 
> Note: I'm unsure whether it makes a noticeable difference as we only output data with 10Hz and the program runs considerably faster than that so for the vast majority of the cases the output would be similar. There could be one print cycle difference.

If I understand you right, you're suggesting possibly placing the boardStatus write code in the `writePin` function or perhaps higher up in the `heaterLoop` function. The former is definitely too low level I would say. The latter is perhaps OK, but definitely results in more distributed code. Also note that so far I've not had to touch the `heatCtrl` module (well, OK, I changed the tabs to spaces...).

I think for the sake of a handful of GPIO reads, in this case we can afford the improvement in readability. Perhaps for something where we are cycle dependent distributing would make sense. 

Final note: The `updateBoardStatus` function is called just before `printCurrentArray` so I don't think there will be any impact on print cycle either way.

LukeWalker-CA:
Waiting on https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/12

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -52,23 +65,59 @@ static CAProtocolCtx caProto =
...
+
+/*!
+** @brief Verbose print of the AC board status
+*/
+static void printAcStatus()
+{
+    static char buf[600] = { 0 };
+    int len = 0;
+
+    if (bsGetStatus() & AC_BOARD_VERSION_ERROR)
```
matias-cphatomics, line 80: Couldn't this be compressed to a call to the statusInfo?

STM32/AC/Core/Src/ACBoard.h
matias-cphatomics: Hm, I get what you are saying and I agree it doesn't really have any major impact. However, if someone actually needed to use this, they would have to update the code anyways, so generally speaking I would leave things out that may or may not be relevant in the future until they become relevant. 

In this instance I don't feel strongly about it though.

matias-cphatomics (approved):

STM32/AC/Core/Src/ACBoard.c
LukeWalker-CA: This was actually removed, but the push didn't happen.

PR #118, opened 2023-10-13
Feature/ac phase control

LukeWalker-CA:
# Phase Control

Phase control works by aligning the start and end of the "on-period" of each heater. If the sum of the "on-period"s overlaps the "PWM period" (e.g. 1000 ms), then it simply loops around, as it is no longer possible to avoid the situation where more than one channel is on at a time
```
Example: 
Heater0=10%, Heater1=30%, Heater2=10%, Heater3=40%
PWM Period |-|-|-|-|-|-|-|-|-|-|
Heater0    |-|
Heater1      |-----|
Heater2            |-|
Heater3              |-------|
            |   |   |    |
            v   v   v    v
           |0|--1--|2|---3---| |

Heater0=20%, Heater1=50%, Heater2=60%, Heater3=60%
PWM Period |-|-|-|-|-|-|-|-|-|-|
Heater0    |---|
Heater1        |---------|
Heater2                  |-----------|
Heater3                              |-----------|
             |      |  |    |
             v      v  v    v
           |-0-|----1----|--2--|
           |--2--|-----3-----| |
```
In example 2, both 0 and 2 must be on at the same time, then 1 and 2, then 1 and 3, then 2 and 3. Finally, 2 will be on for 10% on its own.

# Unit tests

Unit tests are implemented for several heatCtrl.c functions (not all, at time of writing, but more will come). Unit testing framework GoogleTest is used. Its relatively easy to get going with, and has a good [Users Guide](http://google.github.io/googletest/)

### DIFF
--- a/STM32/AC/HeatCtrl/Src/HeatCtrl.c
+++ b/STM32/AC/HeatCtrl/Src/HeatCtrl.c
@@ -1,8 +1,16 @@
 #include <time32.h>
 #include "HeatCtrl.h"
 
+/***************************************************************************************************
+** DEFINES
+***************************************************************************************************/
+
 #define MAX_DURATION ((uint32_t) -1)
 #define MAX_TIMEOUT  60000 // Auto regulation time out from overheat prevention mode
+
+/***************************************************************************************************
+** TYPEDEFS
+***************************************************************************************************/
 
 typedef struct HeatCtrl
 {
@@ -19,8 +27,23 @@ typedef struct HeatCtrl
     uint32_t periodDuration;
 } HeatCtrl;
 
+/***************************************************************************************************
+** PRIVATE VARIABLES
+***************************************************************************************************/
+
 static HeatCtrl heaters[MAX_NO_HEATERS];
 static int noOfHeaters = 0;
+
+/***************************************************************************************************
+** PRIVATE FUNCTION DECLARATIONS
+***************************************************************************************************/
+
+void setPwmPercent(HeatCtrl* ctx, uint32_t pct);
+void updateHeaterPhaseControl();
+
+/***************************************************************************************************
+** FUNCTION DEFINITIONS
+***************************************************************************************************/
 
 HeatCtrl* heatCtrlAdd(StmGpio *heater, StmGpio * button)
 {
@@ -52,8 +75,7 @@ void heaterLoop()
         if (tdiff > pCtrl->periodDuration)
         {
             // Turn of heater since duration is done.
-            pCtrl->pwmPercent = 0;
-            updateHeaterPhaseControl();
+            setPwmPercent(pCtrl, 0);
         }
 
         // Use modules '%' to get the on/off section in PWM period.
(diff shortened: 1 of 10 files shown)


freddyrios:
Some questions/comments on how this runs and how it relates to overall control:

what happens if LoopControl sends changes in the middle of a control period?
what happens if LoopControl resends the expected pwm in the middle of a control period?
on a system with 5 ac boards on the same phase, we will likely be turning on at least 5 heaters at a time
if we run pid control on the LoopControl side, is there an impact on when LoopControl vs. each ac board is running its control period?

STM32/AC/HeatCtrl/Src/HeatCtrl.c
```diff
@@ -126,35 +130,56 @@ void setPWMPin(int pin, int pwmPct, int duration_ms)
...
+**
+** Aligns the start and end of the "on" pulse of all channels (excluding "always on" or "always off"
+** channels) by modifying the "periodBegin" variable for each heater
+*/
+void updateHeaterPhaseControl() 
+{
+    uint32_t period_begin = HAL_GetTick();
+
+    /* Starting at the end and working backwards means all the periodBegins are set BACK in time, 
+    ** instead of forwards. This means the tdiff_u32() function will produce normal results */
```
freddyrios, line 179: might want to add where the tdiff_u32 is being used for better context

freddyrios (commented):
does the build need to be change to run these style of tests? 

e.g. currently build steps make it seem not
image

LukeWalker-CA:
`what happens if LoopControl resends the expected pwm in the middle of a control period?`

Will be re-sent every 2 seconds. Given the PWM period gets re-organised every time (and is 1 second long), this could have quite a severe impact.

Will add an exception so that re-assertions of the same PWM don't reset re-organise the window.

LukeWalker-CA:
After talking with Matias, I realise there will also be an impact on pwmDuration, since the periodBegin of some of the heaters could potentially be set far back in time

- [ ] Remove pwmDuration bug

STM32/AC/HeatCtrl/Src/HeatCtrl.c
LukeWalker-CA: Comment has been removed now

LukeWalker-CA:
> does the build need to be change to run these style of tests?
> 
> e.g. currently build steps make it seem not image

Eventually the build should run these, but this is a scope creep I don't want for this sprint. adding the unit tests is already enough. I'll add it to next sprint - https://github.com/copenhagenatomics/CA_Embedded/issues/119

unit_testing/hello_test.cpp
```diff
@@ -0,0 +1,9 @@
+#include <gtest/gtest.h>
+
+// Demonstrate some basic assertions.
+TEST(HelloTest, BasicAssertions) {
```
matias-cphatomics, line 4: As gtest now seems to be working well, let's remove this before merging.

unit_testing/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+####################################################################################################
+## List of tests to run
+###################################################################################################
+
+enable_testing()
+
+include(GoogleTest)
+
+# HelloTest
+add_executable(hello_test hello_test.cpp)
```
matias-cphatomics, line 38: Let's remove the hello_test before merging

unit_testing/hello_test.cpp
LukeWalker-CA: Good move

PR #121, opened 2023-10-17
Useful DFU script

LukeWalker-CA:
```
usage: dfu.py [-h] [-b BOARD] [-p PORT] [-r] [-m]

Download firmware to board.

options:
  -h, --help            show this help message and exit
  -b BOARD, --board BOARD
                        The board to download to
  -p PORT, --port PORT  The port the board is on, if not already in DFU
  -r, --reset           Reset a board in DFU mode back to normal mode
  -m, --make            Only make, don't download
```
Note: I haven't tested it thoroughly, but I'm using it a bit. Maybe its helpful for you too. Sometimes it works even if minicom is already open on the port (might be worth investigating in more detail, because then it could really speed things up).

Call it from the root of the repo.

### DIFF
--- a/util/dfu.py
+++ b/util/dfu.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+
+import argparse
+import subprocess
+import os
+import serial as s
+from time import sleep
+
+if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


matias-cphatomics (approved):
Beautiful

PR #126, opened 2023-11-27
DC Board Status

LukeWalker-CA:
The DC board now outputs a 32-bit status code (hex) at the end of each data transmission every 10Hz. The 16 MSB bit are defined by the status library (part of systemInfo.c/.h). The DC board specific status code is located in the LSB and are defined as follows:   

| 3 (MSB) | 2 | 1 | 0 (LSB) |
| --- | --- | --- | --- |
| Port 4 On/Off | Port 3 On/Off | Port 2 On/Off | Port 1 On/Off |

The status bits display the port_values, e.g. whether the port has a PWM of more than 0.

The "Status" command displays verbose On/Off status + the current PWM level.

* Fixed a bug where turning the port off didn't trigger the "turnOffPin" function.

@matias-cphatomics Do you think it would be helpful to re-print the "bootMsg" in the output from the "Status" command?


STM32/DC/Core/Src/DCBoard.c
```diff
@@ -285,24 +366,49 @@ static void checkButtonPress()
...
+** Checks the hardware matches this FW version, starts the USB communication and starts the ADC. 
+** Printing is synchronised with ADC, so it must be started in order to print anything over the USB
+** link
+*/
 void DCBoardInit(ADC_HandleTypeDef *_hadc, I2C_HandleTypeDef *_hi2c, WWDG_HandleTypeDef* hwwdg)
 {
-    BoardType board;
-    if (getBoardInfo(&board, NULL) || board != DC_Board)
-        return;
+    setFirmwareBoardType(DC_Board);
```
matias-cphatomics, line 382: Where are these functions defined and what is their purpose? Haven't seen them before and can't find the implementation in the PR.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -191,7 +272,7 @@ static void actuatePins(ActuationInfo actuationInfo)
         return;
     }
 
-    if (actuationInfo.timeOn == -1 && (actuationInfo.percent == 100 || actuationInfo.percent == 0))
+    if (actuationInfo.timeOn == -1000 && (actuationInfo.percent == 100 || actuationInfo.percent == 0))
```
matias-cphatomics, line 275: I'm guessing the update from -1 to -1000 has to do with the bug mentioned in the PR summary. Can you elaborate on the difference here?

matias-cphatomics:
Looks good overall. Just a few comments.

Should implement the same type of test functions for regulating the pin output behaviour as has been done on the AC board? They will probably differ slightly given that it is now an actual PWM pin rather than a "modulated" GPIO pin, but the I think it would be a good add-on. 

Let me know what you think.

matias-cphatomics:
I think it is a good idea to add the bootMsg to the verbose statusInfo output.

STM32/DC/Core/Src/DCBoard.c
LukeWalker-CA: These functions tell systemInfo what board type and oldest supported PCB type the firmware was compiled for.

STM32/DC/Core/Src/DCBoard.c
LukeWalker-CA: Just noticed the same bug is present in the AC board

LukeWalker-CA:
Unit tests to follow in a different PR.

PR #135, opened 2024-01-22
Feature/ac board update

LukeWalker-CA:
* Bug with indefinite times has been fixed (although then removed later, see below)
* AC board "allOn" function, now has a duration added. This also means a duration needs to be added to the CAProtocol "all on" command (See https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/15). "allOff" doesn't require a duration argument as it is always indefinite.
* actuatePins function has been reworked to remove the possibility of turning on any pins without a duration
* adjustPWMDown function now increases duration up to a maximum of 60 secs (down from indefinite), if the PWM must be reduced to reduce board temperature
* updateHeaterPhaseControl now runs asynchronously rather than synchronous to communication (e.g. references from 0, rather than from HAL_GetTick() of whenever it is called) 
* Setting a PWM percent sets a shadow register called pwmNextPct. If pwmNextPct is non-zero at the end of a PWM period, the PWM phases are updated (so the new PWM takes effect at the beginning of the next PWM cycle)
  * Turning on any pin (or all pins) takes effect immediately.


STM32/AC/Core/Src/ACBoard.c
```diff
@@ -143,7 +162,7 @@ static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
     static bool isCalibrationDone = false;
     static int16_t current_calibration[ADC_CHANNELS];
 
-    if (!isComPortOpen()) return;
+    // if (!isComPortOpen()) return;
```
LarsWithCA, line 165: Undo?

unit_testing/AC/AC_tests.cpp
```diff
@@ -7,11 +7,11 @@
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
 
-#include "fake_StmGpio.h"
-#include "fake_stm32xxxx_hal.h"
+// #include "fake_StmGpio.h"
+// #include "fake_stm32xxxx_hal.h"
```
LarsWithCA, line 11: Undo?

unit_testing/CMakeLists.txt
```diff
@@ -35,7 +35,12 @@ enable_testing()
...
\ No newline at end of file
+add_executable(heatCtrl_test AC/heatCtrl_tests.cpp ${LIB}/Util/Src/time32.c fakes/fake_stm32xxxx_hal.cpp fakes/fake_StmGpio.c)
+target_include_directories(heatCtrl_test PRIVATE fakes ${LIB}/Util/Inc ${SRC}/AC/HeatCtrl/Inc ${SRC}/AC/HeatCtrl/Src)
+target_link_libraries(heatCtrl_test GTest::gtest_main gmock_main)
+gtest_discover_tests(heatCtrl_test)
+
+# add_executable(ac_test AC/AC_tests.cpp)
+# target_include_directories(ac_test PRIVATE fakes ${SRC}/AC/Core/Src)
+# target_link_libraries(ac_test GTest::gtest_main gmock_main)
+# gtest_discover_tests(ac_test)
```
LarsWithCA, line 46: Delete?

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -178,57 +197,67 @@ static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
...
-    else if (actuationInfo.timeOn == -1 && actuationInfo.pwmDutyCycle != 0 && actuationInfo.pwmDutyCycle != 100)
+}
+
+static void CAallOn(bool isOn, int duration_ms)
+{
+    if (isOn)
     {
-        // pX on ZZZ%
-        setPWMPin(actuationInfo.pin, actuationInfo.pwmDutyCycle, actuationInfo.timeOn);
+        if(duration_ms < 0)
```
LarsWithCA, line 247: should this be <= ?

matias-cphatomics: +1

matias-cphatomics (approved):

LarsWithCA (approved):

PR #137, opened 2024-02-01
Bugfix/aircon board wrong pinout

LukeWalker-CA:
* Aircon board pcbVersion 1.8 FW was never updated to use the new pinout for the IRED.
* Moved some timer assignments around to achieve the same behaviour, but to also output PWM for the IRED on a second pin, so that either assembly method can be used


LukeWalker-CA:
@matias-cphatomics This has been tested in a few loops, seems to work fine

matias-cphatomics (approved):
Nice and simple way to be backwards compatible. We only need to move the AC related code to the correct branch after which I think it is ready for merging.

LukeWalker-CA:
Closing this due to accidental merge of functionalities. The intended code has been replaced by the following PRs:
* https://github.com/copenhagenatomics/CA_Embedded/pull/141
* https://github.com/copenhagenatomics/CA_Embedded/pull/142

PR #139, opened 2024-02-15
Feature/new dc board pinout

LukeWalker-CA:
* Added 2 new DC outputs
  * Used second timer to PWM these pins - necessitated changing some code to multiplex between timer channels more effectively
* Removed temperature sensor code
* Removed motor driver pins
* Updated pcbversion to be 3.1


STM32/DC/Core/Src/DCBoard.c
```diff
@@ -246,12 +221,13 @@ static void autoOff()
 
 static void handleButtonPress()
 {
-    // Button ports
-    GPIO_TypeDef *button_ports[] = { Btn_2_GPIO_Port, Btn_1_GPIO_Port,
-            Btn_4_GPIO_Port, Btn_3_GPIO_Port };
-    const uint16_t buttonPins[] = { Btn_2_Pin, Btn_1_Pin, Btn_4_Pin, Btn_3_Pin };
+    // Button ports - Button 1 and 2 position is swapped because... insert reason here?
+    GPIO_TypeDef *button_ports[] = { Btn_2_GPIO_Port, Btn_1_GPIO_Port, Btn_4_GPIO_Port,
```
matias-cphatomics, line 225: Agree this swap looks like a bug / bad pin out implementation. The easiest thing is probably just renaming them in the .ioc.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -285,8 +261,32 @@ static void checkButtonPress()
...
+**
+** @param[in] pinNumber Number of DC port to get corresponding timer for
+**
+** @return Address of CCR register, or 0 if an incorrect pinNumber is given
+*/
+static volatile uint32_t* getTimerCCR(int pinNumber)
+{
+    switch (pinNumber)
+    {
+        case 0: return &(TIM4->CCR2);
```
matias-cphatomics, line 275: Can you explain the ordering of the different cases here? 

In the previous implementation, if I'm not mistaken, TIM5 channel 1 through 4 corresponded to pin 1 through 4. In the getTimerCCR function they are now pin 3 through 6. 

Isn't TIM4 pointing to port 5 and 6?

Also it looks like CCR4 and CCR3 are swapped with CCR4 being case 4 and CCR3 being case 5.

matias-cphatomics: So just to be clear I would have imagined them to be listed in the following order:

```
case 0: return &(TIM5->CCR1);
case 1: return &(TIM5->CCR2);
case 2: return &(TIM5->CCR3);
case 3: return &(TIM5->CCR4);
case 4: return &(TIM4->CCR1);
case 5: return &(TIM4->CCR2);
default: return (uint32_t*)0x00000000;
```

With TIM5 pointing to the first 4 ports as they did previously in increasing channel order. And TIM4 pointing to the new ports 5-6 also in increasing order.

STM32/DC/Core/Src/DCBoard.c
LukeWalker-CA: I will rename them in the IOC

STM32/DC/Core/Src/DCBoard.c
LukeWalker-CA: I agree, it would be nice if they're laid out that way, but they aren't. This function essentially maps the pin to the timer CCR register. I guess Anders did some pin swaps when he did the layout. 

Its confusing, but I deliberately moved it all to a single place.

I'll add a comment explaining that the timer layout is correct

STM32/DC/Core/Src/DCBoard.c
matias-cphatomics: I just updated the code to have the order be from 1-6 which matches with the buttons.

STM32/DC/Core/Src/DCBoard.c
matias-cphatomics: @LukeWalker-CA  I just tested with PCB v3.1 that the commands p[1-6] on/off matches with the physical ports. I don't know whether we should then just keep as is, or changed the code to be easier to understand. I don't care to much tbh as it works on the board.

matias-cphatomics: Wait scratch the above. I just read your previous comment properly.

PR #141, opened 2024-02-19
Aircon board fix alterative PR

LukeWalker-CA:
Output IRED PWM on both old and new pins to catch all assembly methods


LukeWalker-CA:
@matias-cphatomics This has been tested in a few loops, seems to work fine

LukeWalker-CA:
> Nice and simple way to be backwards compatible. We only need to move the AC related code to the correct branch after which I think it is ready for merging.

PR #150, opened 2024-04-08
LightController PCB V1.1

matias-cphatomics:
**Updates**
- Change pinout that includes a dedicated white channel for each port
- Change the switching method to be interrupt based since all LEDs are now running as GPIOs rather than PWMs
- Add status code
- Replaced old unittest with gtest unittests (and added more tests)

**Board specific status code**
|15 (MSB) |... | 2| 1| 0 (LSB) |
| ---| --- | --- | --- | --- |
| -|...| LED P3 On/Off| LED P2 On/Off| LED P1 On/Off|

**Main files of interest**
_project_
Core/Src/LightController.c

_tests_
LightController_tests.cpp


LukeWalker-CA:
Note: @matias-cphatomics has updated CA_Embedded_Libaries. He has confirmed on his machine that all other CA_Embedded projects continue to compile with the update.

STM32/LightController/Core/Src/main.c
```diff
@@ -214,15 +208,14 @@ static void MX_TIM2_Init(void)
...
 
   /* USER CODE BEGIN TIM2_Init 1 */
 
   /* USER CODE END TIM2_Init 1 */
   htim2.Instance = TIM2;
-  htim2.Init.Prescaler = 63;
+  htim2.Init.Prescaler = 0;
   htim2.Init.CounterMode = TIM_COUNTERMODE_UP;
-  htim2.Init.Period = 255;
+  htim2.Init.Period = 799;
```
LukeWalker-CA, line 218: Not important, but was considering it might be helpful, particularly for reviewers or people taking over to set the prescaler to the next smallest sensible interval.
E.g. for 20 kHz timer down from 16 MHz, the prescaler would be 15 (so 1 period = 1 us) and the time would be 50.
I think it would be more understandable for reviewing, but also quicker for iterating in development.

Anyway, not crucial, just an idea I had

STM32/LightController/Core/Src/main.c
```diff
@@ -234,170 +227,15 @@ static void MX_TIM2_Init(void)
...
-  HAL_TIM_MspPostInit(&htim2);
-
-}
-
-/**
-  * @brief TIM3 Initialization Function
-  * @param None
-  * @retval None
-  */
-static void MX_TIM3_Init(void)
```
LukeWalker-CA, line 275: What did TIM3 used to do?

STM32/LightController/Core/Src/stm32f4xx_hal_msp.c
```diff
@@ -93,32 +91,13 @@ void HAL_TIM_Base_MspInit(TIM_HandleTypeDef* htim_base)
   /* USER CODE END TIM2_MspInit 0 */
     /* Peripheral clock enable */
     __HAL_RCC_TIM2_CLK_ENABLE();
+    /* TIM2 interrupt Init */
+    HAL_NVIC_SetPriority(TIM2_IRQn, 0, 0);
+    HAL_NVIC_EnableIRQ(TIM2_IRQn);
```
LukeWalker-CA, line 96: The priority is the same as TIM5. I think it would be better if you could work out which was higher priority

unit_testing/LightController/LightController_tests.cpp
```diff
@@ -0,0 +1,432 @@
...
+            }
+
+            // Reset counter
+            if (tickDest == 0){
+                while (tickCounter < TICK_COUNTER_OVERFLOW){
+                    simTick();
+                }
+                tickCounter = 0;
+            }
+        }
```
LukeWalker-CA, line 106: We should move these tick functions (and the USB macros) to a unit testing util file. Maybe next time if you're in a hurry with this PR

unit_testing/LightController/LightController_tests.cpp
```diff
@@ -0,0 +1,432 @@
...
+    auto expectStmNull = [](StmGpio* stm) {
+        EXPECT_EQ(stm->set,    nullptr);
+        EXPECT_EQ(stm->get,    nullptr);
+        EXPECT_EQ(stm->toggle, nullptr);
+    };
+
+    auto expectStmNotNull = [](StmGpio* stm) {
+        EXPECT_NE(stm->set,    nullptr);
+        EXPECT_NE(stm->get,    nullptr);
+        EXPECT_NE(stm->toggle, nullptr);
```
LukeWalker-CA, line 244: I know this is copied from me, but you may want to use ASSERT_NE instead, because if they _are_ null, line 256 will segfault

unit_testing/LightController/LightController_tests.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+    // Wrong hex code (too long)
+    writeLightControllerMessage("p1 13908CC\n");
+    EXPECT_FLUSH_USB(ElementsAre(
+        "\r",
+        "MISREAD: p1 13908CC"
+    ));
+
+    // Wrong hex code (non valid hex characters)
+    writeLightControllerMessage("p1 ACACFM\n");
```
LukeWalker-CA, line 341: I guess a common thing might be to write **0x**123456, should that be allowed?

unit_testing/LightController/LightController_tests.cpp
```diff
@@ -0,0 +1,432 @@
...
+        }
+        EXPECT_FALSE(stmGetGpio(*ChCtrl[i]));
+    } 
+
+    goToTick(171);
+    for(int i = 0; i < LED_CHANNELS*NO_COLORS; i++) EXPECT_FALSE(stmGetGpio(*ChCtrl[i]));
+
+
+    /* --- Turn on p3 white while p2 is running --- */
+    goToTick(0);
```
LukeWalker-CA, line 415: It might be more interesting to try this while still at tick 170/171, or some other number other than 0

LukeWalker-CA (approved):

LukeWalker-CA:
I don't think I saw a check that the status bits are correctly printed when the LED is on (e.g. on every line output)

STM32/LightController/Core/Src/LightController.c
```diff
@@ -76,124 +114,150 @@ bool isInputValid(const char *input, int *channel, unsigned int *rgb)
...
-		portHandle = htim3_;
-		break;
-
-		case 3:
-		portHandle = htim4_;
-		break;
+		rgbwControl[channel*NO_COLORS] 	   = 0;
+		rgbwControl[channel*NO_COLORS + 1] = 0;
+		rgbwControl[channel*NO_COLORS + 2] = 0;
+		rgbwControl[channel*NO_COLORS + 3] = 256;
```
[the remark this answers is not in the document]
matias-cphatomics: We chose to not update the input format with the extra hex value for the white color to not need to different formats to be supported by LoopCode. 

It does limit the functionality of the white color slightly. It could maybe be parsed such that equal numbers on the RGB colors result in that specific on time for the white color such that the intensity can be chosen freely.

STM32/LightController/Core/Src/LightController.c
```diff
@@ -76,124 +114,150 @@ bool isInputValid(const char *input, int *channel, unsigned int *rgb)
...
+}
 
-	// If the user specifies white then turn on white pin separately and turn off
-	// other LED lights
-	for (int i = 1; i <= LED_CHANNELS; i++)
+static void updateLEDs()
+{
+	// Interrupt speed of ledUpdateTimer is 20kHz.
+	// For each 10kHz update the counter incremented until 256 and reset
+	// Hence, the update speed of the led channels are 20kHz/256 ~= 80Hz.
```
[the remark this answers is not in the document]
matias-cphatomics: Yes, it seems sufficient.

STM32/LightController/Core/Src/main.c
matias-cphatomics: TIM2,3 and 4 used to be PWM timers.

STM32/LightController/Core/Src/LightController.c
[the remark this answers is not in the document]
matias-cphatomics: The 256 is a bug - I've updated it to 255.

STM32/LightController/Core/Src/LightController.c
```diff
@@ -76,124 +114,150 @@ bool isInputValid(const char *input, int *channel, unsigned int *rgb)
...
+		bsSetError(BS_VERSION_ERROR_Msk);
+        return;
+    }
+
+	pcbVersion ver;
+	if (getPcbVersion(&ver) || ver.major != 1 || ver.minor < 1)
+	{
+		bsSetError(BS_VERSION_ERROR_Msk);
+		return;
+	}
```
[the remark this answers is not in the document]
matias-cphatomics: I've added the extra check so that it is not possible to control a PCB where the pinout does not match the SW.

STM32/LightController/Core/Src/stm32f4xx_hal_msp.c
matias-cphatomics: Fair point. I've set the switching timer to have higher priority.

unit_testing/LightController/LightController_tests.cpp
matias-cphatomics: From discussion the USB macros have been moved into fake_USBprint.h, but the tick functions differ so they remain as they are.

unit_testing/LightController/LightController_tests.cpp
matias-cphatomics: I'm having a hard time actually getting this working. Since loopcode never uses the 0x prefix I will leave it a later exercise.

PR #153, opened 2024-04-09
UnitTest: Use fakes/stubs in CA_Embedded_Libraries location

matias-cphatomics:
Setup CA_Embedded to run gtest unittests.

**Changes**
- Removed "master" CMakeLists.txt and added individual CMakeLists.txt for each testing module.
- Updated the unitTests.py to run the CMakeLists.txt in the chosen subdirectory specified in -R input
- Updated workflow to run gtest unittests rather than the previous raw .cpp style.

**TODO**
We should update what happens in the ifdef UNIT_TESTING part of systeminfo.c targeting some of the build variables in CA_Embedded_Libraries to avoid compiler warnings when running tests. Also possibly ignore warnings that trigger because we are running tests on 64-bit machines rather than the 32-bit target system.


.github/workflows/stm32build.yml
```diff
@@ -44,31 +44,27 @@ jobs:
...
-        cmake CMakeLists.txt
-        make
-        output=$(./CA_UnitTesting)
-        echo $output
-        if [[ $output =~ "failed" ]]; then
-          echo "Test failed"
-          exit 1
-        fi
+        cd unit_testing
+        python unitTests.py -R ${{ matrix.package }}
```
LukeWalker-CA, line 67: Have you verified this "fails" if a unit test fails?

unit_testing/unitTests.py
```diff
@@ -10,8 +10,9 @@
     parser.add_argument('-v', '--verbose', action="store_true", help='Maximum output')
     args = parser.parse_args()
 
-    subprocess.run("cmake -S . -B build", shell=True)
-    subprocess.run("cmake --build build", shell=True)
+    dir=f"{args.regex}"
```
LukeWalker-CA, line 13: As discussed:
* Separate directory and regex concepts (-R and -d args perhaps)
* If no directory argument, scan subfolders and execute each one (but not the build folder)

.github/workflows/stm32build.yml
LukeWalker-CA, line 1: As discussed:
* Update YAML to run unit tests for all modules if CA_Embedded_Libraries is updated

unit_testing/unitTests.py
LukeWalker-CA: This is because Regex is useful for running a single test out of a module, as well as selecting the module out from many modules

matias-cphatomics: - Separate directory and regex concepts using -D and -R are now implemented.

- If no directory argument is passed the then the unitTests.py scans the immediate subfolders (hence excluding nested build folders) and execute each one.

.github/workflows/stm32build.yml
matias-cphatomics: As discussed this will be part of a separate PR.

unit_testing/unitTests.py
```diff
@@ -3,22 +3,40 @@
 import argparse
 import subprocess
 import os
+import sys
+
+def run_tests_in_subdirectory(dir, regex, verbose):
```
LukeWalker-CA, line 8: Neither regex nor verbose are used in this function. They should be appended to run_str

LukeWalker-CA (changes requested):

LukeWalker-CA (approved):

PR #155, opened 2024-04-16
Added remote hosts download to DFU

LukeWalker-CA:
* Also added make -j8 option for faster build

### DIFF
--- a/util/dfu.py
+++ b/util/dfu.py
@@ -13,6 +13,7 @@ if __name__ == "__main__":
     parser.add_argument('-d', '--dfu', action="store_true", help='Use instead of -p if the board is in DFU mode')
     parser.add_argument('-R', '--remote', help='The hostname for the port (e.g. loop name)')
     parser.add_argument('-P', '--password', help='The password for the host')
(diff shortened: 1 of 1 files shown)


util/dfu.py
```diff
@@ -27,26 +29,58 @@
...
-                        sleep(0.1)
-                except s.SerialException:
-                    pass
-                finally:
-                    brd.close()
-
-                sleep(0.5)
-
-            subprocess.run(f"dfu-util -a 0 -D build/{args.board}.bin -s 0x08000000:leave", shell=True)
+            subprocess.run(f"make -j8", shell=True, check=True)
```
matias-cphatomics, line 32: We have seen issues on windows machines when compiling with the -j8 tag. Because the compilation is spread over different cores in some scenarios part of the project that is dependent on the githash.h is built before the githash.h is produced. This means the build will fail. 

Maybe it should be yet another option flag.

matias-cphatomics (approved):

util/dfu.py
LukeWalker-CA: Done

PR #157, opened 2024-05-27
Bugfix/dc board status

LukeWalker-CA:
* Fixed board status bug (I think)
* Unit tests
* Fixed bug where input ports were not valid, causing a segfault


STM32/DC/Core/Src/DCBoard.c
```diff
@@ -63,6 +64,14 @@ static uint32_t actuationStart[ACTUATIONPORTS] = { 0 };
 static bool port_state[ACTUATIONPORTS] = { 0 };
 static uint32_t ccr_states[ACTUATIONPORTS] = { 0 };
 
+// Button ports - Button 1 and 2 position is swapped because... insert reason here?
```
matias-cphatomics, line 67: I think this comment should be removed

unit_testing/DC/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+
+####################################################################################################
+## List of tests to run
+###################################################################################################
+
+enable_testing()
+
+include(GoogleTest)
+
+# AC tests
```
matias-cphatomics, line 44: DC :=)

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,197 @@
...
+        "Board Status:\r", 
+        "The board is operating normally.\r", 
+        "Port 0: On: 0, PWM percent: 0\r",
+        "Port 1: On: 1, PWM percent: 999\r",
+        "Port 2: On: 0, PWM percent: 0\r",
+        "Port 3: On: 1, PWM percent: 999\r",
+        "Port 4: On: 0, PWM percent: 0\r",
+        "Port 5: On: 1, PWM percent: 999\r",
+        "\r", 
+        "End of board status. \r"
```
matias-cphatomics, line 195: Let's have part of the test check that the error bit is set if the 24V supply is missing.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -129,7 +153,7 @@ static void printResult(int16_t *pBuffer, int noOfChannels, int noOfSamples)
         return;
     }
```
matias-cphatomics, line 154: On line 152 I think the format specifier should be changed from `0x%x` to `0x%08x` for consistency.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -138,68 +162,54 @@ static void printResult(int16_t *pBuffer, int noOfChannels, int noOfSamples)
...
 {
-    pinWrite(pinNumber, RESET);
     actuationDuration[pinNumber] = 0;
-    port_state[pinNumber] = 0;
-    ccr_states[pinNumber] = *(getTimerCCR(pinNumber));
+    port_state[pinNumber] &= ~PORT_STATE_ON_SW;
+    ccr_states[pinNumber] = TURNOFFPWM;
 }
 
 typedef struct ActuationInfo
```
matias-cphatomics, line 215: Comment on line 217 should be updated to follow the new 6 port pin out.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -213,33 +223,36 @@ static void actuatePins(ActuationInfo actuationInfo)
 {
     if (!actuationInfo.isInputValid)
     {
+        HALundefined("Invalid input");
+        return;
+    }
+    else if (actuationInfo.pin < 0 || actuationInfo.pin >= ACTUATIONPORTS) {
```
matias-cphatomics, line 229: It seems odd to me that this isn't being captured in the isInputValid check above.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -213,33 +223,36 @@ static void actuatePins(ActuationInfo actuationInfo)
...
-    else if (actuationInfo.timeOn == -1000 && actuationInfo.percent != 0 && actuationInfo.percent != 100)
+    /* Note: conditions in comments met by implication */
+    else if (actuationInfo.timeOn == -1000) /* && actuationInfo.percent != 0 && actuationInfo.percent != 100 */
     {
         // pX on ZZZ%
         int pwmState = (actuationInfo.percent * MAX_PWM) / 100;
         setPWMPin(actuationInfo.pin, pwmState, 0);
     }
-    else if (actuationInfo.timeOn != -1000 && actuationInfo.percent != 100)
+    else if (actuationInfo.timeOn != -1000 && actuationInfo.percent != 0) /* && actuationInfo.percent != 100 */
```
matias-cphatomics, line 255: On line 270 add an empty line for format.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -274,30 +287,22 @@ static void autoOff()
...
-    // Button ports - Button 1 and 2 position is swapped because... insert reason here?
-    GPIO_TypeDef *button_ports[] = { Btn_1_GPIO_Port, Btn_2_GPIO_Port, Btn_3_GPIO_Port, 
-                                     Btn_4_GPIO_Port, Btn_5_GPIO_Port, Btn_6_GPIO_Port};
-    const uint16_t buttonPins[] = { Btn_1_Pin, Btn_2_Pin, Btn_3_Pin, 
-                                    Btn_4_Pin, Btn_5_Pin, Btn_6_Pin };
-
     for (int i = 0; i < ACTUATIONPORTS; i++)
     {
-        if (HAL_GPIO_ReadPin(button_ports[i], buttonPins[i]) == 0)
+        if (stmGetGpio(buttonGpio[i]) == 0)
```
matias-cphatomics, line 292: I think we should add a comment here that a button press gives a logical 0 and vice versa for clarity.

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -336,6 +341,35 @@ static volatile uint32_t* getTimerCCR(int pinNumber)
     }
 }
 
+/*!
```
matias-cphatomics, line 344: In line 331 the comment should be updated to V3.2 and of course test that that is in fact true for a V3.2 :=)

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -336,6 +341,35 @@ static volatile uint32_t* getTimerCCR(int pinNumber)
...
+** @brief Initialises GPIO in the system
+*/
+static void gpioInit () {
+    for(int i = 0; i < 6; i++) {
```
matias-cphatomics, line 348: Change 6 to ACTUATION_PORTS

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -346,7 +380,7 @@ void DCBoardInit(ADC_HandleTypeDef *_hadc, WWDG_HandleTypeDef* hwwdg)
     setFirmwareBoardVersion((pcbVersion){3, 1});
 
     /* Always allow for DFU also if programmed on non-matching board or PCB version. */
-    initCAProtocol(&caProto, usb_cdc_rx);
+    initCAProtocol(&caProto, usbRx);
     
     BoardType board;
     if (getBoardInfo(&board, NULL) || board != DC_Board)
```
matias-cphatomics, line 386: Line 391 change to PCB V3.1

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -383,4 +419,5 @@ void DCBoardLoop(const char* bootMsg)
     // Turn off pins if they have run for requested time
     autoOff();
     checkButtonPress();
+    handlePorts();
```
matias-cphatomics, line 422: I'm not quite sure whether this could pose a problem or not, but given it is an actuation board I think we should be careful with running handlePorts if there is a SW mismatch. 

The timers are started in main.c and initialised to a safe state, but if for whatever reason a command is sent to them they could begin to actuate even if there is a sw mismatch where the behaviour is unknown.

One way to tackle this would be to move the initialisation of the timers to the DCBoardInit function and then only initialise them if there is no sw mismatch.

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+/* Real supporting units */
+#include "ADCmonitor.c"
+#include "CAProtocol.c"
+#include "CAProtocolStm.c"
+
+/* UUT */
+#include "DCBoard.c"
+
+using ::testing::AnyOf;
```
matias-cphatomics, line 23: AnyOf, AllOf and IsEmpty are not used.

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,506 @@
...
+            /* TODO: Make this use a define for the board release, so that tests fail if the someone
+            ** forgets to update the version numbers */
+            BoardInfo bi = {
+                .v2 = {
+                    .otpVersion = OTP_VERSION_2,
+                    .boardType  = DC_Board,
+                    .subBoardType = 0,
+                    .reserved = {0},
+                    .pcbVersion = {
+                        .major = 3,
```
matias-cphatomics, line 54: Should we go to the new format of the PCB version checking using pcbversion.sh and so on?

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+        void writeDcMessage(const char * msg)
+        {
+            hostUSBprintf(msg);
+            DCBoardLoop(bootMsg);
+        }
+
+        void simTick(int numTicks = 1)
+        {
+            for(int i = tickCounter + 1; i <= tickCounter + numTicks; i++)
```
matias-cphatomics, line 73: What is the reason for `tickCounter + 1` here? I tried removing it and the unit tests still pass.

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,506 @@
...
+    dcSetup();
+
+    /* Basic test, was everything OK?  */
+    EXPECT_FALSE(bsGetStatus() && BS_VERSION_ERROR_Msk);
+
+    /* This should force a print on the USB bus */
+    goToTick(100);
+
+    /* Check the printout is correct */
+    EXPECT_READ_USB(Contains("0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0x00000000"));
```
matias-cphatomics, line 141: Let's do a negative test here as well.

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,506 @@
...
+                if(j != i) {
+                    ASSERT_EQ(*getTimerCCR(j-1), 0) << "j = " << j;
+                }
+                else {
+                    ASSERT_EQ(*getTimerCCR(j-1), 999) << "j = " << j;
+                }
+            }
+        }
+
+        writeDcMessage("all off\n");
```
matias-cphatomics, line 270: Would it make sense to see that the port that had just been set now has a CCR of 0 again?

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+/* Grey box - uses getTimerCCR() */
+TEST_F(DCBoard, portsNoTimeout) 
+{
+    dcSetup();
+
+    goToTick(1);
+    
+    char cmd[100] = {0};
+    for(int i = 0; i < 8; i++) {
```
matias-cphatomics, line 248: Replace 8 with `<=ACTUATION_PORTS + 1`

unit_testing/DC/DC_tests.cpp
matias-cphatomics: Wait it is already checked in the next loop iteration of course.

unit_testing/DC/DC_tests.cpp
```diff
@@ -0,0 +1,506 @@
...
+        /* Timer port should turn off 1 second after the start of the test */
+        simTick(700);
+        ASSERT_EQ(*getTimerCCR(i), 999) << "i = " << i;   
+        simTick();
+        ASSERT_EQ(*getTimerCCR(i), 0) << "i = " << i;   
+    }
+}   
+
+/* Grey box - uses getTimerCCR() */
+TEST_F(DCBoard, onboardButtonsOffDuring)
```
matias-cphatomics, line 442: Being petty here but During what :question: -  could use a update to the test name.

matias-cphatomics (approved):

STM32/DC/Core/Src/DCBoard.c
LukeWalker-CA: There is no check. I'll remove that variable.

STM32/DC/Core/Src/DCBoard.c
LukeWalker-CA: Added an if statement to prevent this

unit_testing/DC/DC_tests.cpp
LukeWalker-CA: TLDR: Performance.

Added comment:
```
            /* tickCounter indicates the current tick. forceTick(now) doesn't make any sense, so 
            ** start from the next tick */

unit_testing/DC/DC_tests.cpp
LukeWalker-CA: Just so you know, YOU are responsible for this:
image

PR #52, opened 2022-03-24
new board names

thomasjamatomics:

PR #164, opened 2024-06-18
Feature/ac fault info

LukeWalker-CA:
* Also see: https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/52
* Added fault handlers to AC board
  * In stm32f4xx_it.c, for each fault handler, there is a call to the FAULT_HANDLER macro.
  * The FAULT_HANDLER macro records the stack frame and a selection of registers in flash.
  * After the FAULT_HANDLER has been called, reset the chip
  * Next time "Serial" is called (e.g. next time LoopControl starts up) the fault info is printed to the USB and cleared.
  * Call this function from somewhere in the code to test:
```
static int illegal_instruction_execution(void) {
  int (*bad_instruction)(void) = (int (*)()) 0xE0000000;
  return bad_instruction();
}
```
* In order to enable writing to flash, the watchdog configuration had to be changed - now it matches pressure board
* Updated unit tests
* Updated to pcbversion.h


STM32/AC/Core/Src/ACBoard.c
```diff
@@ -85,6 +88,16 @@ static CAProtocolCtx caProto =
 ** PRIVATE FUNCTIONS
 ***************************************************************************************************/
 
+static void printAcHeader()
+{
+    CAPrintHeader();
+
+    /* If a serious fault occurs that requires a reset */
```
matias-cphatomics, line 95: /* If a serious fault occurred that required a reset */

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -156,9 +169,6 @@ static double ADCtoTemperature(double adc_val)
 WWDG_HandleTypeDef* hwwdg_ = NULL;
 static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
 {
-    // Update window watchdog. Will trigger reset outside window: <90ms && >110ms.
-    HAL_WWDG_Refresh(hwwdg_);
```
matias-cphatomics, line 160: So the iwdg crashes the program when storing the fault info in flash so it's removed. However, I don't see why we wouldn't keep on using the wwdg as an actual window wdg keeping the previous timer settings. 

The wwdg is being disabled when running flash so should be fine.

STM32/AC/Core/Src/main.c
```diff
@@ -257,34 +251,6 @@ static void MX_ADC1_Init(void)
...
-
-  /* USER CODE BEGIN IWDG_Init 0 */
-
-  /* USER CODE END IWDG_Init 0 */
-
-  /* USER CODE BEGIN IWDG_Init 1 */
-
-  /* USER CODE END IWDG_Init 1 */
-  hiwdg.Instance = IWDG;
-  hiwdg.Init.Prescaler = IWDG_PRESCALER_4;
```
matias-cphatomics, line 276: Did you try changing the IWDG prescalar value before removing it? The `IWDG_PRESCALER_4` is the strictest option. 

I am almost convinced that choosing a higher prescalar would solve the IWDG triggering during flashing.

matias-cphatomics (approved):
Looks good - however, if possible, I think we should keep both watch dogs on. I think we might be able to simply adjust the iwdg prescalar, but if not I'm happy with this.

STM32/AC/Core/Src/ACBoard.c
LukeWalker-CA: Actually the IWDG isn't the problem. The IWDG can be set slow enough that it won't interfere with the flash operation

The problem seems to be that the print timing (and therefore watchdog reset) is driven off the ADC DMA. The DMA seems to be reset (or perhaps it keeps going, but the interrupt flag can't be set while flashing?) during the flash write operation.

PR #178, opened 2024-11-07
AC Ten channel PCB V1.0

matias-cphatomics:
**AC Ten Channel PCB V1.0**

 - Functionality similar to AC board with 10 output channels rather than 4 (Status bits, Number of ADC channels and print out changed accordingly)
 - 1-wire interface to DS18B20 temperature sensor implemented (previously TMP35 measured using ADC channel)
 - Removed unused button interface
 - Update to the libraries to support the new board type in this [PR](https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/53)
 - Added basic unit tests using a fake of the DS18B20 in this [PR](https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/59)
 - Points to the STM32F4 drivers in embedded libraries
 - Memory layout is such that the calibration data is placed in the smaller FLASH sectors and the program is stored in sectors 4 and 5.


STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+    }
+    else 
+    {
+        HALundefined(input);
+    }
+}
+
+static void GpioInit()
+{
+    const int NO_PORTS = AC_TEN_CH_NUM_PORTS;
```
TimotheDodinCA, line 126: Why do you redefine the number of ports and not use directly `AC_TEN_CH_NUM_PORTS` ?

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+        heatCtrlAdd(&heaterPorts[i]);
+    }
+
+    stmGpioInit(&fanCtrl, FAN_CTRL_GPIO_Port, FAN_CTRL_Pin, STM_GPIO_OUTPUT);
+}
+
+static double ADCtoCurrent(double adc_val)
+{
+    static float CURRENT_SCALAR = 0.013138;
+    static float CURRENT_BIAS = -0.01;
```
TimotheDodinCA, line 144: You could define them as `const`

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+        else if((HAL_GetTick() - port_close_time) >= USB_COMMS_TIMEOUT_MS)
+        {
+            allOff();
+        }
+        return;
+    }
+    port_close_time = 0;
+
+    /* If the version is incorrect, there is no point printing data or doing maths */
+    if (bsGetStatus() & BS_VERSION_ERROR_Msk)
```
TimotheDodinCA, line 173: Use `bsGetField()` instead

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+            // finding the average of each channel array to subtract from the readings
+            current_calibration[i] = -ADCMean(pData, i);
+        }
+        isCalibrationDone = true;
+    }
+
+    // Set bias for each channel.
+    for (int i = 0; i < noOfChannels; i++)
+    {
+        // Go from channel 1 since 0 is temperature.
```
TimotheDodinCA, line 192: It seems that you do it for temperature as well now so you can remove this comment

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+            allOff();
+        }
+        return;
+    }
+    port_close_time = 0;
+
+    /* If the version is incorrect, there is no point printing data or doing maths */
+    if (bsGetStatus() & BS_VERSION_ERROR_Msk)
+    {
+        USBnprintf("0x%08" PRIx32, bsGetStatus());
```
TimotheDodinCA, line 175: Is this better than `USBnprintf("0x%x", bsGetStatus());` ?

STM32/ACTenChannel/Core/Src/ACTenChannel.c
matias-cphatomics: Yeah the comment does not make sense anymore. Previously, the temperature was on ADC channel 0, but it is now measured via the DS18B20. Anyway, I'll remove the comment.

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+static void updateBoardStatus();
+
+/***************************************************************************************************
+** PRIVATE OBJECTS
+***************************************************************************************************/
+
+/* GPIO settings. */
+StmGpio heaterPorts[AC_TEN_CH_NUM_PORTS];
+static StmGpio fanCtrl;
+static StmGpio DQ1;
```
TimotheDodinCA, line 65: What does DQ mean?

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+
+    /* Don't initialise any outputs or act on them if the board isn't correct */
+    if(boardSetup(ACTenChannel, (pcbVersion){BREAKING_MAJOR, BREAKING_MINOR}) == -1)
+    {
+        return;
+    }
+
+    static int16_t ADCBuffer[ADC_CHANNELS * ADC_CHANNEL_BUF_SIZE * 2]; // array for all ADC readings, filled by DMA.
+    ADCMonitorInit(hadc, ADCBuffer, sizeof(ADCBuffer)/sizeof(int16_t));
+    HAL_TIM_Base_Start(htim);
```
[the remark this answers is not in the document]
matias-cphatomics: It should be `HAL_TIM_Base_Start`. It should only be `HAL_TIM_Base_Start_IT` when you want the timer to invoke the `HAL_TIM_PeriodElapsedCallback`, but here it is only used as a timer for the ADC conversions. Hence, it should be this way in the AC project as well.

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,395 @@
...
+** @brief Loop function called repeatedly throughout
+** 
+** * Responds to user input
+** * Checks for ADC buffer switches (the ADC sample rate is synchronised with USB print rate - the 
+**   USB print rate should be 10 Hz, so every 400 ADC samples the buffer switches).
+** * Runs the closed loop control system for board temperature and PWMs the heaters as per user 
+**   input
+*/
+void ACTenChannelLoop(const char *bootMsg)
+{
```
TimotheDodinCA, line 387: Maybe you can add the new stack trace print implemented recently on the AC board:
if(CAhandleUserInputs(&caProto, bootMsg)) {
        /* If a serious fault that required a reset occurs, print the stack trace immediately upon boot */
        if(printFaultInfo()) {
            clearFaultInfo();
            fhSaveDeposit();
        }
    };``

STM32/ACTenChannel/Core/Src/DS18B20.c
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+/*!
+** @brief Receives a byte from the DS18B20
+*/
+static uint8_t readData()
+{
+    uint8_t value = 0;
+    setPinInput();
```
TimotheDodinCA, line 100: I don't think this line is needed as you make it an output on the next line

STM32/ACTenChannel/Core/Src/DS18B20.c
```diff
@@ -0,0 +1,201 @@
...
+            acquiring = 0;
+            return temp; 
+        }
+
+        writeData(0xCC); // skip ROM
+        writeData(0x44); // start temperature acquisition
+
+        resetTime = HAL_GetTick();
+    } 
+    else if (acquiring && ((HAL_GetTick() - resetTime) >= 800))
```
TimotheDodinCA, line 169: I think it's better to use `tdiff_u32()` from `time32.h` to take care of the overflow

STM32/ACTenChannel/STM32F401CCUx_FLASH.ld
```diff
@@ -0,0 +1,189 @@
...
+_estack = ORIGIN(RAM) + LENGTH(RAM);    /* end of RAM */
+/* Generate a link error if heap and stack don't fit into RAM */
+_Min_Heap_Size = 0x200;      /* required amount of heap  */
+_Min_Stack_Size = 0x400; /* required amount of stack */
+
+/* Specify the memory areas */
+MEMORY
+{
+RAM (xrw)      : ORIGIN = 0x20000000, LENGTH = 64K
+FLASH (rx)      : ORIGIN = 0x8000000, LENGTH = 256K
```
TimotheDodinCA, line 65: Don't you want to update the flash at the same time?

STM32/ACTenChannel/Core/Src/ACTenChannel.c
matias-cphatomics: DQ is the data line of the DS18B20 chip according to the datasheet :)

STM32/ACTenChannel/Core/Src/DS18B20.c
matias-cphatomics: All the other implementations I've seen has had this seemingly redundant call. I will try to remove it and see if it still works :=)

STM32/ACTenChannel/Core/Src/DS18B20.c
```diff
@@ -0,0 +1,201 @@
...
+        {
+            return temp; 
+        }
+
+        writeData(0xCC); // skip ROM
+        writeData(0xBE); // read temperature from scratch pad
+
+        uint8_t adc_low = readData();
+        uint8_t adc_high = readData();
+        adc = (adc_high << 8) | adc_low;
```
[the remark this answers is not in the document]
matias-cphatomics: Roger

STM32/ACTenChannel/Core/Src/DS18B20.c
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/*!
+** @brief Returns after us microseconds
+** @note The DS18B20_tim updates with 1MHz
+*/
+static void delay_us(uint32_t us)
+{
+    __HAL_TIM_SET_COUNTER(DS18B20_tim, 0);
+    while (DS18B20_tim->Instance->CNT < us);
```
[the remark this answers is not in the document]
matias-cphatomics: Maybe we define a `const MAX_US` that is equal (or maybe a few ticks higher) than the max used for the DS18B20. Could also be entirely separate as in max 1 ms using HAL_GetTick where it brakes if more than a ms has passed.

STM32/ACTenChannel/Core/Src/ACTenChannel.c
```diff
@@ -0,0 +1,393 @@
...
+** Loops keeping the board cool using the attached fan. If the board gets too hot, starts to reduce
+** the power on all ports by reducing the PWM
+*/
+static void heatSinkLoop()
+{
+    static unsigned long previous = 0;
+    // Turn on fan if temp > 55 and turn of when temp < 50.
+    if (heatSinkTemperature <= MAX_TEMPERATURE)
+    {
+        if (heatSinkTemperature < 50 && !isFanForceOn)
```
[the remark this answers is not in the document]
matias-cphatomics: There may be one in the future, but it is probably a bit silly to keep there while it doesn't have one. It is copied more or less directly from the AC implementation, so should be easy to include at another point. Will remove it :=)

PR #180, opened 2024-11-13
Remove unused .NET

matias-cphatomics:
This C# project was used for a short while as a visualisation tool for the SaltFlow board. It hasn't been used for 2+ years.


LukeWalker-CA (changes requested):
Maybe you should remove the .NET references from the .gitignore

LukeWalker-CA (approved):

PR #185, opened 2025-01-03
AC: PCB V6.5

matias-cphatomics:
- Pinout change
- Added 3 extra temperature channels + powerStatus GPIO
- Updated unit tests to match new print format
- Added powerStatus to status bits
- Updated main clock speed to 84MHz from 16MHz and updated timer pre-scalars (and removed unnecessary timer interrupt)


STM32/AC/Core/Src/ACBoard.c
```diff
@@ -76,7 +80,7 @@ static CAProtocolCtx caProto =
         .printHeader = printAcHeader,
         .printStatus = printAcStatus,
```
TimotheDodinCA, line 81: You could define a prinfStatusDef function :) to make it easier to use the status code

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -207,18 +213,29 @@ static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
     }
 
     // Set bias for each channel.
-    for (int i = 1; i < noOfChannels; i++)
+    for (int i = 0; i < NUM_CURRENT_CHANNELS; i++)
     {
         // Go from channel 1 since 0 is temperature.
```
TimotheDodinCA, line 218: You can remove this comment

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -207,18 +213,29 @@ static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
...
+        // Temperature channels are 4-7
+        heatSinkTemperatures[i] = ADCtoTemperature(ADCMean(pData, i+4));
+        if (heatSinkTemperatures[i] > maxTemp)
+        {
+            maxTemp = heatSinkTemperatures[i];
+        }
+    }
+    heatSinkMaxTemp = maxTemp;
+
+    USBnprintf("%.4f, %.4f, %.4f, %.4f, %.2f, %.2f, %.2f, %.2f, 0x%x", ADCtoCurrent(ADCrms(pData, 0)), ADCtoCurrent(ADCrms(pData, 1)),
```
TimotheDodinCA, line 234: We started to use `0x%08 PRIx32` on other boards to make sure the status code is always composed of the same number of characters

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -207,18 +213,29 @@ static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
...
+        heatSinkTemperatures[i] = ADCtoTemperature(ADCMean(pData, i+4));
+        if (heatSinkTemperatures[i] > maxTemp)
+        {
+            maxTemp = heatSinkTemperatures[i];
+        }
+    }
+    heatSinkMaxTemp = maxTemp;
+
+    USBnprintf("%.4f, %.4f, %.4f, %.4f, %.2f, %.2f, %.2f, %.2f, 0x%x", ADCtoCurrent(ADCrms(pData, 0)), ADCtoCurrent(ADCrms(pData, 1)), 
+                                                                       ADCtoCurrent(ADCrms(pData, 2)), ADCtoCurrent(ADCrms(pData, 3)),
```
TimotheDodinCA, line 235: Do we want to implement some check on the current so that if a current on on port is too high it will be turned off (to detect short circuit for example before it becomes warm)

TimotheDodinCA (approved):

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -207,18 +213,29 @@ static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
...
         ADCSetOffset(pData, current_calibration[i], i);
     }
 
-    heatSinkTemperature = ADCtoTemperature(ADCMean(pData, 0));
-    USBnprintf("%.4f, %.4f, %.4f, %.4f, %.2f, 0x%x", ADCtoCurrent(ADCrms(pData, 1)),
-            ADCtoCurrent(ADCrms(pData, 2)), ADCtoCurrent(ADCrms(pData, 3)),
-            ADCtoCurrent(ADCrms(pData, 4)), 
-            heatSinkTemperature,
-            bsGetStatus());
+    double maxTemp = 0;
```
LukeWalker-CA, line 222: What if all the temperatures are sub-zero? Maybe this should start at -DBL_MAX

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -207,18 +213,29 @@ static void printCurrentArray(int16_t *pData, int noOfChannels, int noOfSamples)
...
+    for (int i = 0; i < NUM_TEMP_CHANNELS; i++)
+    {
+        // Temperature channels are 4-7
+        heatSinkTemperatures[i] = ADCtoTemperature(ADCMean(pData, i+4));
+        if (heatSinkTemperatures[i] > maxTemp)
+        {
+            maxTemp = heatSinkTemperatures[i];
+        }
+    }
+    heatSinkMaxTemp = maxTemp;
```
LukeWalker-CA, line 232: Is it possible to turn the channels on, then disconnect the USB so this is never re-calculated? I guess there are hopefully some timeouts that would prevent anything getting too hot, but should we rely on them?

LukeWalker-CA (approved):
Generally looks very good :)

STM32/AC/Core/Src/ACBoard.c
matias-cphatomics: From discussion: The computation of the heat sink temperatures will be moved to its own function. It will be called if the USB port is closed as an addition to where it's computed now. This ensures that a loss of communication with board will not cause the temperature to stall and the logic that uses it other places will still be able to run.

STM32/AC/Core/Src/ACBoard.c
matias-cphatomics: The fuses on each port of the AC board is at 10A, so if there is anything above that we have a physical safety mechanism. Currents below that is allowed and should in principal be able to run indefinitely without the board overheating. In the case of overheating we have the heatSinkLoop that turns on the fan and lowers the PWMs if the board is above its max temperature. 

In addition there is a loopcontrol plugin that measures the current draw on each port before operation. This is of course not a board feature.

For the cases I foresee, I think this is sufficient. If there is another case I didn't catch let me know :=)

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -158,10 +193,24 @@ static double ADCtoCurrent(double adc_val)
...
+    for (int i = 0; i < NUM_TEMP_CHANNELS; i++)
+    {
+        heatSinkTemperatures[i] = ADCtoTemperature(ADCMean(pData, i+NUM_CURRENT_CHANNELS));
+        if (heatSinkTemperatures[i] > maxTemp)
+        {
+            maxTemp = heatSinkTemperatures[i];
+        }
+    }
+    heatSinkMaxTemp = maxTemp;
 }
```
LukeWalker-CA, line 214: I know you didn't introduce this in this PR, but what is the watchdog handle doing on line 216?

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -352,12 +401,20 @@ static void heatSinkLoop()
...
+
     stmGetGpio(fanCtrl) ? bsSetField(AC_BOARD_PORT_x_STATUS_Msk(0)) : bsClearField(AC_BOARD_PORT_x_STATUS_Msk(0));
     for(int i = 0; i < AC_BOARD_NUM_PORTS; i++)
     {
         stmGetGpio(heaterPorts[i].heater) ? bsSetField(AC_BOARD_PORT_x_STATUS_Msk(i+1)) : bsClearField(AC_BOARD_PORT_x_STATUS_Msk(i+1));
     }
 
+    /* Heavily averaged signal of the last 1000 samples to smooth out large dips
+    ** The gpio input is not supposed to change value generally so should be fine to use a large filter
+    ** Note: When power is toggled the status code changes within one print cycle */
```
LukeWalker-CA, line 414: Have I got it right that the filter is ~100 seconds long? That seems like overkill to fix the glitch problem. Surely 1 second would be enough? Does it mean it would take ~50 seconds to register loss of power?

LukeWalker-CA (approved):

STM32/AC/Core/Src/ACBoard.c
matias-cphatomics: I'll remove it.

I think it is a "left over" from when the wwdg was used with a [90-110]ms window before introducing all the flash functionalities.

STM32/AC/Core/Src/ACBoard.c
matias-cphatomics: That would be a misunderstanding. 

The `updateBoardStatus()` function is called at every iteration of the main loop. So I don't know exactly how long the 1000 iterations take, but as I wrote in the comment the status code updates within one print cycle i.e. no delay from a pihub's perspective.

I tried using smaller filter lengths e.g. 10, 50 and 100, but there would always be a periodic glitch in the status code.

matias-cphatomics: I'm open to do something smarter, if you have an idea though :=)

LukeWalker-CA: Gotcha - now I understand the comment. Seems fine to me

PR #190, opened 2025-01-29
AC: Channel layout + Max temp handling

matias-cphatomics:
Changes:

- Reverse the ADC rank of the temperature channels to match physical ports
- Remove adjustPWM (used when board temperature exceeds max). Instead a time limit of 10 seconds on time per request is added (previously there was no limit). This way if the board exceeds max temperature the board will run for a max of 10 seconds in this state before turning off channels. It will ignore all future on requests until the temperature has fallen below the max temperature. If >10 seconds is requested the board turns on for 10 sec and a status bit is set notifying the user of the truncated time output.


STM32/AC/Core/Src/ACBoard.c
```diff
@@ -375,17 +384,13 @@ static void heatSinkLoop()
 
         bsClearField(BS_OVER_TEMPERATURE_Msk);
     }
-    else // Safety mode to avoid overheating of the board
+    else 
     {
+        /* Board is running above max temperature 
+        ** NOTE: As the max on time per request is 10 seconds, it is deemed safe
+        **       to let the board run until the next timeout (<=10 sec) occurs.
+        **       While the board is above max temperature it ignores any new requests. */
```
LukeWalker-CA, line 392: Just to be clear, the line 341 is what prevents new requests?
```
if (heatSinkMaxTemp > MAX_TEMPERATURE && percent > pwmPercent)
```
And the assumption is that after the duration timeout, pwmPercent will be 0?

Is there a bug here that the heat protection could be semi-bypassed by the following sequence?
```
p0 on 100% 10
Wait 9.5 seconds
p0 on 99% 10
Wait 9.5 seconds
etc...
```

STM32/AC/Makefile
```diff
@@ -27,7 +27,7 @@ TARGET = AC
 # building variables
 ######################################
 # debug build?
-DEBUG = 1
+DEBUG = 0
```
LukeWalker-CA, line 30: Noice

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -375,17 +384,13 @@ static void heatSinkLoop()
...
         stmSetGpio(fanCtrl, true);
```
LukeWalker-CA, line 393: Should you remove the fan code?

LukeWalker-CA (commented):
Overall looks straightforwards - potential bug

STM32/AC/Core/Src/ACBoard.c
matias-cphatomics: The board doesn't seem to be quite able to run full throttle without a fan at this point and the GPIO is still connected to a fan pin.

As it is fairly harmless and doesn't interfere with other functionality I would prefer to keep it for a little while to come.

TimotheDodinCA (approved):
Looks good to me!

PR #191, opened 2025-02-03
Update DC input protocol

matias-cphatomics:
- The inputCAProtocol now redirects AC / DC specific handler code to the `caProto.undefined()` function which in turn calls the inputhandler defined in the PR below.
- Update of CA_Embedded_Libraries can be found in this [PR](https://github.com/copenhagenatomics/CA_Embedded_Libraries/actions/runs/13073231940/job/36479235207)
 - Update of the build script as the version we were using of the upload-artifact action has been deprecated.

### DIFF
--- a/STM32/DC/Core/Src/DCBoard.c
+++ b/STM32/DC/Core/Src/DCBoard.c
@@ -91,7 +91,7 @@ static StmGpio sense24v;
 */
 static void DCInputHandler(const char* input)
 {
-    ACDCInputHandler(&caProto);
+    ACDCInputHandler(&caProto, input);
 }
 
 /*!
--- a/STM32/DC/Core/Src/DCBoard.c
+++ b/STM32/DC/Core/Src/DCBoard.c
@@ -14,6 +14,7 @@
 #include "ADCMonitor.h"
 #include "CAProtocol.h"
 #include "CAProtocolStm.h"
+#include "CAProtocolBoard.h"
 #include "time32.h"
 #include "StmGpio.h"
 #include "pcbversion.h"
@@ -39,6 +40,7 @@
 ** PRIVATE FUNCTION DECLARATIONS
 ***************************************************************************************************/
 
+static void DCInputHandler(const char* input);
 static void CAallOn(bool isOn, int duration_ms);
 static void CAportState(int port, bool state, int percent, int duration);
 static volatile uint32_t* getTimerCCR(int pinNumber);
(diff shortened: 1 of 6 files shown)


STM32/DC/Core/Src/DCBoard.c
```diff
@@ -53,7 +55,7 @@ static void handlePorts();
 
 static CAProtocolCtx caProto =
 {
-        .undefined = HALundefined,
+        .undefined = DCInputHandler,
         .printHeader = CAPrintHeader,
         .printStatus = printDcStatus,
         .jumpToBootLoader = HALJumpToBootloader,
```
LukeWalker-CA, line 61: Given that allOn and portState are no longer used as part of the general protocol, should they be moved to their own structure in the CAProtocolBoard file?

LukeWalker-CA (approved):
Simple

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -77,13 +79,21 @@ static GPIO_TypeDef *button_ports[] = { Btn_1_GPIO_Port, Btn_2_GPIO_Port, Btn_3_
...
 static StmGpio sense24v;
 
 /***************************************************************************************************
 ** PRIVATE FUNCTION DEFINITIONS
 ***************************************************************************************************/
 
+/*!
+** @brief Call command handler for DC board
+*/
+static void DCInputHandler(const char* input)
```
LukeWalker-CA, line 92: Is this an opportunity to implement the timeout behaviour that AC has on DC as well?

LukeWalker-CA (approved):

STM32/DC/Makefile
```diff
@@ -51,6 +51,7 @@ C_SOURCES =  \
 ../../CA_Embedded_Libraries/STM32/USBprint/Src/USBprint.c \
 ../../CA_Embedded_Libraries/STM32/Util/Src/CAProtocol.c \
 ../../CA_Embedded_Libraries/STM32/Util/Src/CAProtocolStm.c \
+../../CA_Embedded_Libraries/STM32/Util/Src/CAProtocolBoard.c \
```
TimotheDodinCA, line 54: You should add it in the .extSettings file as well

STM32/DC/Core/Src/DCBoard.c
TimotheDodinCA: Maybe we should add this feature in the library as it seems interesting for all actuation boards

TimotheDodinCA (approved):
Good!

PR #196, opened 2025-02-14
LightController: Library update

matias-cphatomics:
- Update to latest CA_Embedded_Libraries where the CAProtocol has been updated so that the command `pX ffffff` is now valid


TimotheDodinCA (approved):
👍

PR #199, opened 2025-02-18
Bugfix/ac heater race

LukeWalker-CA:
* Fixed a race condition where turning on a pin after setting PWM period could result in the PWM period taking priority.
* Added unit test that catches the failure before the fix is added


matias-cphatomics (approved):
Good catch. 

Out of curiosity, did you see this behaviour in a system or did you just spot the potential error case? This PR would only update the AC PCB V6.5 which is not yet on the shelves, so if this was a reported bug we should also do a patch version.

LukeWalker-CA:
Freddy reported this with low priority. I thinking patching earlier versions might be a good idea. Which versions should I patch?

matias-cphatomics:
Patch versions:

- PCB V6.4 (breaking version is 6.4 or maybe 6.3, but no releases exist for 6.3)
- PCB V6.2 (breaking version looks like it might be 5.9)

The PCB V6.4 is the one most systems have installed now, but according to Anders there may be some 6.2s out there.

LukeWalker-CA:
Patches for others done, ready for complete review

PR #200, opened 2025-02-19
Race condition bugfix

LukeWalker-CA:
* Removed turn[On|Off]Pin race condition
* Editted YML for a (hopefully) more fluid release experience


LukeWalker-CA:
Wrong source branch

PR #213, opened 2025-02-26
DC: PCB v3.2

matias-cphatomics:


STM32/DC/Core/Src/DCBoard.c
```diff
@@ -43,6 +43,9 @@
 
 #define USB_COMMS_TIMEOUT_MS 5000
 
+#define UNDER_VOLTAGE_ADC 150
```
TimotheDodinCA, line 46: You could put some comment saying which voltage it is approximately

STM32/DC/Core/Src/DCBoard.c
```diff
@@ -200,6 +213,8 @@ static void printResult(int16_t *pBuffer, int noOfChannels, int noOfSamples)
         return;
     }
 
+    inputVoltageADC = ADCMean(pBuffer, 6);
+
```
TimotheDodinCA, line 217: Here (or somewhere else) you can use bsSetVoltage()

TimotheDodinCA (approved):
Good job!
Concerning your two questions:
- I think it's better to keep the voltage as an ADC value, given the variability between the boards. It could be misinterpreted in my opinion
- Yes we need to change the fact that the under/over voltage error is port specific, but I'm fine if we do it later and you open an issue

STM32/DC/stm32f401ccux_flash.ld
```diff
@@ -183,7 +183,6 @@ SECTIONS
     libgcc.a ( * )
   }
 
-  .ARM.attributes 0 : { *(.ARM.attributes) }
```
LukeWalker-CA, line 186: important?

unit_testing/DC/DC_tests.cpp
```diff
@@ -115,7 +121,7 @@ class DCBoard: public ::testing::Test
                 .reserved = {0},
                 .pcbVersion = {
                     .major = 3,
-                    .minor = 1
+                    .minor = 2
```
LukeWalker-CA, line 124: Use "LATEST_MAJOR and LATEST_MINOR" here instead of 3 and 2

These unit tests are a bit out of date - possibly consider updating to the newest format?

unit_testing/DC/DC_tests.cpp
```diff
@@ -237,18 +243,19 @@ TEST_F(DCBoard, printStatus)
 TEST_F(DCBoard, status24v) 
 {
     dcSetup();
-    stmSetGpio(sense24v, false);
+
+    /* Fill buffer with ADC value resembling under voltage */
+    setADCChannelBuffer(inputVoltageChannelIdx, 0);
```
LukeWalker-CA, line 248: Why not test at the limit, instead of at 0?

unit_testing/DC/DC_tests.cpp
```diff
@@ -260,9 +267,14 @@ TEST_F(DCBoard, status24v)
...
     ));
+
+    /* Fill buffer with ADC value resembling over voltage */
+    setADCChannelBuffer(inputVoltageChannelIdx, 1200);
+    goToTick(400);
+    EXPECT_FLUSH_USB(Contains("0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0xb0000000"));
 }
 
-/* Grey box - uses getTimerCCR() */
+// /* Grey box - uses getTimerCCR() */
```
LukeWalker-CA, line 277: Gotta make sure that comment is commented out

unit_testing/DC/DC_tests.cpp
LukeWalker-CA: I'm not going to insist on it, but I think a good unit test would actually loop through all the values between 0 and the limit to check undervoltage, and the same for overvoltage.

LukeWalker-CA (approved):
No major concerns, except some interest about why the EOC flag changed and the ARM attributes left.

STM32/DC/Core/Src/main.c
```diff
@@ -212,11 +213,11 @@ static void MX_ADC1_Init(void)
...
   hadc1.Init.DiscontinuousConvMode = DISABLE;
   hadc1.Init.ExternalTrigConvEdge = ADC_EXTERNALTRIGCONVEDGE_RISING;
-  hadc1.Init.ExternalTrigConv = ADC_EXTERNALTRIGCONV_T2_TRGO;
+  hadc1.Init.ExternalTrigConv = ADC_EXTERNALTRIGCONV_T3_TRGO;
   hadc1.Init.DataAlign = ADC_DATAALIGN_RIGHT;
-  hadc1.Init.NbrOfConversion = 6;
+  hadc1.Init.NbrOfConversion = 7;
   hadc1.Init.DMAContinuousRequests = ENABLE;
-  hadc1.Init.EOCSelection = ADC_EOC_SINGLE_CONV;
+  hadc1.Init.EOCSelection = ADC_EOC_SEQ_CONV;
```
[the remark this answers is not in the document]
matias-cphatomics: I set it to `ADC_EOC_SEQ_CONV`  thinking that I had somehow set it to `ADC_EOC_SINGLE_CONV` by mistake. In all other projects it is `ADC_EOC_SINGLE_CONV`.

It turns out that it doesn't have an impact because the DMA reads and transfers the content of the `ADC_DR` register every time it is updated. The end of conversion / end of sequence is relevant when the user handles incoming ADC samples via software. 

The behaviour is described in section 11.8 in the reference manual, if interested.

STM32/DC/stm32f401ccux_flash.ld
matias-cphatomics: The `.ARM.attributes` is only a metadata section i.e. not a functional section - apparently mainly used for tools like `objdump` for analysis purposes. 

As far as I could read for some compilers you will get a warning if it is not present, but since it is omitted by ST in their new firmware packages I doubt it will cause issues.

LukeWalker-CA:
Forgot to answer your questions:

* I think its fine to convert it as a channel (with hefty notes about accuracy). I don't believe it would be used without consulting us anyway, and it could be a useful diagnostic in case of some failure.
* I think its fine to do another PR (or maybe just raise an issue). I don't get entirely what you mean, as it is just a single flag. Are you saying its just the wording of the warning?

PR #218, opened 2025-03-04
ZrO2: Move to CA_Embedded

matias-cphatomics:
- [x]  Moved project
- [x]  Moved unittests
- [x]  Updated workflow.yml
- [ ]  **Make draft tag-release to ensure correct release numbering** (Draft tag should be `ZirconiumOxide-2.7.0`)


TimotheDodinCA (approved):

LukeWalker-CA (changes requested):
I don't think Oxygen is supposed to be open-sourced

matias-cphatomics:
> I don't think Oxygen is supposed to be open-sourced

You're right - I'll close these two PRs.

PR #223, opened 2025-03-21
Pressure: PCB V3.4

matias-cphatomics:
**Changes:**
- Updated pin out
- Removed sub-board type functionality (not used in any systems anymore)
- Updated to `pcbversion.h` format
- Updated unit tests
- Added `StatusDef` command
- Updated memory layout - **because its a new breaking pcb version** -  to use less space for calibration data.
- Updated to new voltage range
- Formatting done in last commit so it is easy to remove modifications in github

**Note:** Please review this [PR](https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/79) first


STM32/Pressure/Core/Src/pressure.c
```diff
@@ -28,26 +40,14 @@ static float volts[ADC_CHANNELS];       // port voltage readings
...
-static void calibrateSensorOrBoard(int noOfCalibrations, const CACalibration* calibrations);
-static void calibrationInfoRW(bool write);
-static void logMode(int port);
-
-// Local variables
 static CAProtocolCtx caProto =
 {
     .undefined = HALundefined,
-    .printHeader = printHeader, // Need this since calibrations should be included.
+    .printHeader = printHeader,
```
TimotheDodinCA, line 50: Can you add a printStatusDef ? You ca put it as NULL if no board specific status bit is defined.

TimotheDodinCA (approved):
All good

TimotheDodinCA:
You can close this issue then [220](https://github.com/copenhagenatomics/CA_Embedded/issues/220)

AndersJorgensenCA:
[Tweaks needed](https://github.com/copenhagenatomics/CA_Embedded/issues/224)

STM32/Pressure/Core/Src/pressure.c
```diff
@@ -1,281 +1,315 @@
...
+ */
+static void ADCtoPressure(float *adcMeans, int noOfChannels) {
+    /*
+    cal.sensorCalVal[channel*2]     - Scale assuming [0V, 5V] range
+    VOLTAGE_SCALING                 - To go to real [0V, 5.112V] range
+    cal.sensorCalVal[channel*2+1]   - Pressure bias
+    */
+    for (int channel = 0; channel < noOfChannels; channel++) {
+        pressure[channel] = adcMeans[channel] * VOLTAGE_SCALING * cal.sensorCalVal[channel * 2] +
+                            cal.sensorCalVal[channel * 2 + 1];
```
LukeWalker-CA, line 187: Has this been checked end-to-end? Something feels wrong here (scaling factor). Is this because LoopControl provides a sensor calibration that assumes the board does 0-5V?

STM32/Pressure/Core/Src/pressure.c
```diff
@@ -1,281 +1,315 @@
...
+         *  Hence, we divide by MAX_VCC_IN (=5.49V) rather than MAX_VIN
+         */
 
-static void printPorts(float *portValues)
-{
-    USBnprintf("%0.6f, %0.6f, %0.6f, %0.6f, %0.6f, %0.6f, 0x%x",
-                *portValues, *(portValues+1), *(portValues+2), *(portValues+3),
-                *(portValues+4), *(portValues+5), bsGetStatus());
+        // Current measurement across shunt resistor without voltage divider
+        if (cal.measurementType[channel]) {
```
LukeWalker-CA, line 206: If I were writing this, I'd have some defines for measurementType
```
MEASUREMENT_TYPE_CURRENT == cal.measurementType[channel]
```
which avoids ambiguity (I'd forgotten by here if current was 1 or 0, so had to go elsewhere in the code to check)

LukeWalker-CA (approved):
I think it all looks good, a few minor points

For future reference (its probably too integrate to do anything about now) LoopControl should really provide a calibration of voltage --> pressure rather than ADC --> Pressure.

TimotheDodinCA:
> I think it all looks good, a few minor points
> 
> For future reference (its probably too integrate to do anything about now) LoopControl should really provide a calibration of voltage --> pressure rather than ADC --> Pressure.

Yes I think the same. It makes no sense that they have to give a calibration from ADC to pressure.

STM32/Pressure/Core/Src/pressure.c
TimotheDodinCA: Yes it's exactly the reason why it is weird.
The range 0-5V is hardcoded in loop control and it sends a calibration value in bar/ADCunit...
This means also that the previous pressure (previous versions) was not read perfectly as the previous exact range was more 0-4.95V...

TimotheDodinCA: And @AndersJorgensenCA tested this morning that the pressure readings makes sense (2mBar difference compared to previous version with same sensor), so good enough.
I believe that the difference comes from the previous range being not exactly 0-5V.

STM32/Pressure/Core/Src/pressure.c
TimotheDodinCA: Yes I will do that

PR #226, opened 2025-04-14
Revert Flash Memory Layout 

LukeWalker-CA:
* Changes to linker scripts to move calibration area to the end of the flash.

Note: Current cannot fit in the flash if the CAL is moved to the end. But LoopControl always sends the calibration when it is turned on, so it is not an immediate problem


TimotheDodinCA (approved):
Did you test it on the boards?
It's a bit annoying not to be able to do it on the Current board for 1.8kB...

LukeWalker-CA:
No, I will do that before I merge

LukeWalker-CA:
Tested and works fine

PR #232, opened 2025-06-12
Bugfix/dc 3.1 misreads

LukeWalker-CA:
### Bug Description

A bug occurs with a low frequency where the following text will be printed in the log (~1-2 times per hour, e.g. once every 10000 commands)
```
[uploader33] Unexpected board response 'MISREAD: p1\r'
```

### Analysis

The root cause of the error has not been located yet. It is very sensitive to the timing of commands, and also the configuration of the stack. Disabling optimisations appears to remove the bug. These things make debugging difficult and time consuming. 

The bug has been traced to occur during the processing of the "getArgs" function. It occurs regardless of whether the function is interrupted or not (during testing the function has been bracketed by __disable_irqs()). Given that getArgs is essentially a wrapper around strtok, it seems likely that the bug is something to do with the strtok function. strtok is an opaque library function.

### Mitigation

The bug should be considered active until the root cause is identified and understood. However removing the strtok function and replacing it with a similar behaviour either fixes or masks the bug, reducing unnecessary warning messages for other teams.

Library pull here: https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/89

### DIFF
--- a/.github/workflows/stm32build.yml
+++ b/.github/workflows/stm32build.yml
@@ -2,11 +2,11 @@ name: C/C++ stm32 build for the open-source embedded software
 
 on:
   push:
-    branches: [ 'main' ]
-    paths: ['STM32/**', '.github/workflows/stm32build.yml']
+    branches: [ 'patch/dc-3.1' ]
+    paths: ['STM32/DC/**', '.github/workflows/stm32build.yml']
   pull_request:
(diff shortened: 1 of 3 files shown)


LukeWalker-CA:
DC3.2 doesn't seem to exhibit the bug, so I think it is unnecessary to make a release for DC3.2

PR #237, opened 2025-07-31
Tachometer V3.4

LukeWalker-CA:
### Primary Changes

* Updated Tachometer to work on pressure boards V3.1 - V3.4

### Secondary Changes

* Updated unit tests
* Added Tachometer workspace file for better VSCODE integration
* Library PR here: https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/94

### Testing

Tested on PressureBoard V3.1, 3.2 and 3.4 (3.3 not in circulation)


STM32/Tachometer/Core/Src/tachometer.c
```diff
@@ -159,6 +168,80 @@ static void resetFlow()
...
+    GPIO_InitStruct.Pull = GPIO_NOPULL;
+    HAL_GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    if(ver.major == 3 && ver.minor >= 3) {
+        __HAL_RCC_GPIOB_CLK_ENABLE();
+        GPIO_InitStruct.Pin = GPIO_PIN_1;
+        HAL_GPIO_Init(GPIOB, &GPIO_InitStruct);
+    }
+
+    if(ver.major == 3 && ver.minor == 1) {
```
TimotheDodinCA, line 211: Am I missing something or those lines are the same as previously?

STM32/Tachometer/Core/Src/tachometer.c
```diff
@@ -159,6 +168,80 @@ static void resetFlow()
...
+    else if(ver.major == 3 && ver.minor == 2) {
+        GPIO_InitStruct.Pin = GPIO_PIN_0|GPIO_PIN_1|GPIO_PIN_2|GPIO_PIN_3|GPIO_PIN_4|GPIO_PIN_5;
+        SET_GPIO_PINS(gpio_pins, GPIO_PIN_5, GPIO_PIN_4, GPIO_PIN_3, GPIO_PIN_2, GPIO_PIN_1, GPIO_PIN_0);
+    }
+    else if(ver.major == 3 && ver.minor >= 3) {
+        GPIO_InitStruct.Pin = GPIO_PIN_2|GPIO_PIN_3|GPIO_PIN_4|GPIO_PIN_6|GPIO_PIN_7;
+        SET_GPIO_PINS(gpio_pins, GPIO_PIN_1, GPIO_PIN_7, GPIO_PIN_6, GPIO_PIN_4, GPIO_PIN_3, GPIO_PIN_2);
+    }
+    /* Enable interrupts */
+    if(ver.minor <= 2) {
```
TimotheDodinCA, line 224: You don't check the major here?
Also does it means for minor > 2 its only has 5 interrupts and not 6?

STM32/Tachometer/STM32F401XX_FLASH.ld
TimotheDodinCA, line 1: Lots of changes in here, I guess because it is the newest FW version?

TimotheDodinCA (approved):
Nice to use the same FW some different versions!

STM32/Tachometer/Core/Src/tachometer.c
LukeWalker-CA: It cannot get to this position if major != 3.

Yes EXT5-EXT9 all share the same IRQ line. The HAL works out which EXTi was actually called in SW. So for minor > 2 2 of the lines share an IRQ

STM32/Tachometer/STM32F401XX_FLASH.ld
LukeWalker-CA: Indeed - 6 years worth of changes.

If you check though, its mainly whitespace, READONLY and alignment, none of which I consider to be problematic

PR #240, opened 2025-09-03
Bugfix/usb interrupt ac 6.2

LukeWalker-CA:
### Primary Changes

* Updated AC board to latest version of the libraries. Required minor changes to ACBoard code
* Within library, usb_cdc_fops.c:usb_cdc_transmit has a bug fixed where is was possible for the USB interrupt to become permanently disabled

### Testing

- [x] unit tests
- [x] Benchtop tests


AndersJorgensenCA:
Why did the light controller have ~2 file changes, but this has ~120?

AndersJorgensenCA (approved):
Fair, I will approve it, but some more testing is probably a good idea.

PR #99, opened 2023-04-14
Testing monorepo

NicolasMichaelCPHATM:

PR #245, opened 2025-10-02
Dev/analogInput

LukeWalker-CA:
### Primary Changes

* New code for AnalogInput board. Loosely based on PressureBoard code, but has significant changes:
  * Commands to change measurement range and voltage output
  * Basic uptime support
  * More straightforwards sensor calibration method. The Pressure board converts to volts/current, and the calibration converts from that unit to the sensor unit
  * Attempts to connect to digital potentiometers after an error
* See Library PR: https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/99

### Testing

- [x] Unit testing
- [x] Alpha desk testing
- [ ] Beta testing


STM32/AnalogInput/Core/Src/analog_input.c
```diff
@@ -0,0 +1,530 @@
...
+static void analogInputCommandHandler(const char* input) {
+    unsigned int channel = 0;
+    float volt_range = 0;
+    if (sscanf(input, "p%d inmax %f", &channel, &volt_range) == 2) {
+        /* Only for channels within range and set to measure voltage. Current measurement is forced
+        * to 20 mA max. */
+        if(channel >= 1 && 
+           channel <= NO_CALIBRATION_CHANNELS && 
+           cal.measurementType[channel - 1] == 0) {
+            setDigipotWiper(&measure_pots[channel - 1], channel - 1, measureVoltageToDigipotIdx(volt_range), false);
```
TimotheDodinCA, line 167: Maybe verify that volt_range is a "valid number", for example, if it is a negative number, then the potentiometer will go to the highest value I think. I would prefer that it prints a message and/or doesn't take into account the command

STM32/AnalogInput/Core/Src/analog_input.c
```diff
@@ -0,0 +1,530 @@
...
+        }
+        else {
+            HALundefined(input);
+        }
+    }
+    else if (sscanf(input, "p%d volt %f", &channel, &volt_range) == 2) {
+        if(channel >= 1 && channel <= NO_CALIBRATION_CHANNELS) {
+            setDigipotWiper(&power_pots[channel - 1], channel - 1, powerVoltageToDigipotIdx(volt_range), true);
```
TimotheDodinCA, line 175: Same here

STM32/AnalogInput/Core/Src/analog_input.c
```diff
@@ -0,0 +1,530 @@
...
+            volts[channel] = adcScaled * MAX_28V_IN;
+        }
+        else { /* channel == ADC_CHANNEL_VBUS */
+            volts[channel] = adcScaled * MAX_VBUS_IN;
+        }
+
+        /* If doing current measurement, convert voltage to current through shunt resistor */
+        if(cal.measurementType[channel]) {
+            /* 160 Ohm shunt resistor */
+            volts[channel] = AMPS_TO_MILLIAMPS * volts[channel] / 160.0;
```
TimotheDodinCA, line 318: Define 160 in a variable

STM32/AnalogInput/Core/Src/analog_input.c
TimotheDodinCA, line 1: So LoopControl will be responsible for sending the voltage range to the board everytime?

STM32/AnalogInput/Core/Src/calibration.c
```diff
@@ -0,0 +1,205 @@
...
+            continue;
+        }
+
+        // Make sure the measurement type has been explicitly set
+        if (calibrations[i].threshold != 0 && calibrations[i].threshold != 1) {
+            continue;
+        }
+
+        const int channel                  = calibrations[i].port - 1;
+        cal->sensorCalVal[channel * 2]     = calibrations[i].alpha;
```
TimotheDodinCA, line 154: Way easier than previously because no more unit conversions, but LoopControl will have to modify the calibration units

STM32/AnalogInput/Core/Src/analog_input.c
TimotheDodinCA: Can it be a problem if a sensor is powered externally and outputs 0-24V for example. Then before the calibration, will the ADC just saturate and give maximum value, or can it damage something?

TimotheDodinCA (approved):
Nice! I wrote a few comments / questions
It was quite difficult to know which code was new / old compared to the pressure board, but no choice I guess

STM32/AnalogInput/Core/Src/analog_input.c
LukeWalker-CA: Yes, loopcontrol will do that. At the moment, I always want the board to boot in 5V range (same as pressure board), otherwise I'm worried sensors will get broken.

If the sensor is powered externally and outputs 24V, the op-amp will saturate and get +10 °C hotter, so no biggie

STM32/AnalogInput/Core/Src/calibration.c
LukeWalker-CA: Yes already spoken to Lars & Freddy. Fortunately since we already agreed AnalogInput will be a "new" board, its not difficult for them to implement the new calibration method

STM32/AnalogInput/Core/Src/analog_input.c
```diff
@@ -468,6 +480,29 @@ static void forceCurrentMeasurementRange() {
     }
 }
 
+/*!
+** @brief Gradually update the digipot wipers to their target position
+*/
+static void updateDigipotWipers() {
```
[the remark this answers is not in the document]
LukeWalker-CA: Yes - the output SMPS doesn't seem entirely stable under no-load conditions, but slower ramping fixes it

PR #246, opened 2025-10-16
Frequency and RoCoF measurements

TimotheDodinCA:
# Frequency and RoCoF measurements

## General changes

- Update to new ST FW package
- New frequency meaurement technique
- Now measures the rate of change as well
- Removal of CMSIS files as not used anymore. I kept the flashing method into 2 files, so the updated boards keep their calibration, but I don't think it matters as LoopControl sends the calibration to the board every time
- Unit test update

**Question:** As I'm updating the FW, should I add the UpTime feature? I don't think it is really useful on this board, but why not, if in the long run we want it on all the boards

## Frequency measurement improvement
**Objective:**
 Get a resolution of around 0.1 Hz displayed every 10 Hz

**Previously:**
FFT calculated with 1024 long buffer at 1024 Hz sampling frequency -> 1 Hz resolution calculated every second
To get 0.1 Hz resolution with an FFT, 10s of buffered data are needed (no matter the sampling frequency) -> Too long

**New method:**
Using a Phase Lock Loop (PLL). This method estimates the phase error between an internal virtual oscillator (cos) and the ADC sample. It then feeds this error (which is normalized) through a PI controller, which has been tuned experimentally. The PI controller outputs the estimated frequency. The phase is then computed by integrating the frequency.
The error is the quadrature component computed by multiplying the ADC sample with the sin of the internal oscillator. (I hope it makes more sense when looking at the code)

## Rate of Change of Frequency (RoCoF) measurement
- Calculated as the derivative of the filtered frequency
- Low pass filtered to limit noise influence
- In the same loop as the PLL

**Tested on synRM pump, I will test it on induction pump before merging.**


STM32/Current/Core/Src/CurrentApp.c
```diff
@@ -25,12 +24,59 @@
...
+} PhaseData_t;
+
 /***************************************************************************************************
 ** PRIVATE FUNCTION DECLARATIONS
 ***************************************************************************************************/
 
 static void CurrentPrintHeader();
 
+static void pllReset(PLL_t *pll);
+static void pllStep(PLL_t *pll, float x);
```
LukeWalker-CA: Should the PLL code be in a separate module?

STM32/Current/Core/Src/CurrentApp.c
```diff
@@ -135,32 +152,92 @@ static void CurrentPrintHeader()
...
+    pll->omega = output;
+
+    // Filtering
+    // Absolute value as PLL can lock on opposite side
+    previousOmegaFilt = pll->omegaFilt;
+    pll->omegaFilt = (1.0 - PLL_ALPHA) * pll->omegaFilt + PLL_ALPHA * fabsf(pll->omega);
+    pll->theta += pll->omega;
+
+    // Wrapping
+    while (pll->theta >= 2.0 * M_PI) {
```
LukeWalker-CA: Could use fmod? More standard

LukeWalker-CA (commented):
Perhaps some links to how the code should function would be helpful

TimotheDodinCA:
> Perhaps some links to how the code should function would be helpful

Is this helpful? (I'm adding it to the file?

/*
    PLL (Phase Lock Loop) principle
    -
    The goal is a make the input signal match with an internal oscillator
    -
    Input signal:
    where A is the signal amplitude and ϕ its phase
    x(t) = A cos(wt + ϕ)
    -
    In phase and quadrature components (comparison between signal and internal oscillator):
    where θ is the internal oscillator phase (not constant)
    inPhase = x(t) cos(θ) = A cos(wt + ϕ) cos(θ) = 0.5 A [ cos(θ − ωt − ϕ) + cos(θ + ωt + ϕ) ]
    quad    = x(t) sin(θ) = A cos(wt + ϕ) sin(θ) = 0.5 A [ sin(θ − ωt − ϕ) + sin(θ + ωt + ϕ) ]
    -
    Low pass filtering
    The second term (with θ + ωt + ϕ) is at twice the signal frequency so it is filtered away
    (θ changes at each step to follow the input signal)
    It remains:
    where Δ is the phase error
    inPhase = 0.5 A cos(θ − ωt − ϕ) = 0.5 A cos(Δ)
    quad    = 0.5 A sin(θ − ωt − ϕ) = 0.5 A sin(Δ)
    -
    If the phase error Δ is brought to 0, it means:
    inPhase -> 0.5 A
    quad    -> 0
    -
    So quad is given as input to a PI controller which output the estimated input frequency
    θ is then obtained by integration
    -
    The signal amplitude A can be computed:
    amp = 2 √[inPhase² + quad²]
        = 2 √[(0.5 A cos(Δ))² + (0.5 A sin(Δ))²]
        = 2 ⋅ 0.5 ⋅ A √[(cos(Δ))² + (sin(Δ))²]
        = A
    */

STM32/Current/Core/Src/pll.c
```diff
@@ -0,0 +1,138 @@
...
+    pll->omega       = 0.0;
+    pll->omegaFilt   = 0.0;
+    pll->rocof       = 0.0;
+    pll->theta       = 1.0;  // Different from 0 so it can starts again
+}
+
+/*
+PLL (Phase Lock Loop) principle
+-
+The goal is a make the input signal match with an internal oscillator
```
LukeWalker-CA: This is a slightly odd phrasing, where I believe the goal is to make the internal oscillator match with the input signal

LukeWalker-CA (approved):
I understand whats happening now, and your new comment goes some way to explaining it. Code seems good!

STM32/Current/Core/Src/pll.c
```diff
@@ -0,0 +1,138 @@
+/*!
+ * @file    pll.c
+ * @brief   This file contains the implementation of a Phase Lock Loop
+ * @date    07/11/2025
+ * @author  Timothé Dodin
```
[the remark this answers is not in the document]
TimotheDodinCA: What's the best between putting the octave script in the repo or linking to a drive?

PR #248, opened 2025-10-22
Temperature: IWDG fix

TimotheDodinCA:
- IWDG threshold changed from 0.8s to 2s
- I had to update the FW package because the unit tests wouldn't run
- StatusDef addition
- Cleaning to make the code more readable
- Removal of duplicate / now useless files (from cubeIDE)

Tested on a few boards (PCB 6.1 and 6.0)

### DIFF
--- a/STM32/Temperature/Core/Inc/Temperature.h
+++ b/STM32/Temperature/Core/Inc/Temperature.h
@@ -18,36 +18,38 @@
 ***************************************************************************************************/
 
 // Extern value defined in .ld linker script
-extern uint32_t _FlashAddrCal;   // Starting address of calibration values in FLASH
-#define FLASH_ADDR_CAL ((uintptr_t) &_FlashAddrCal)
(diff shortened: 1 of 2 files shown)


STM32/Temperature/Core/Src/main.c
```diff
@@ -215,7 +216,7 @@ static void MX_IWDG_Init(void)
   /* USER CODE END IWDG_Init 1 */
   hiwdg.Instance = IWDG;
   hiwdg.Init.Prescaler = IWDG_PRESCALER_16;
-  hiwdg.Init.Reload = 2499;
+  hiwdg.Init.Reload = 4095;
```
LukeWalker-CA: Can you double check the maths here? I think either 0.8 or 2 s is wrong.

Maybe there should be a comment in the main loop about how long the IWDG is supposed to be?

TimotheDodinCA: Previously it was:
( (2499 * 16) / ( 32 kHz )) = 1.2495 s
So yes it was not 0.8s

PR #255, opened 2026-02-03
Update for temperature PCB V6.2

TimotheDodinCA:
Changes:

- Changes MCU pin layout according to PCB version
- Unused pins test1/test2/led removed from IOC file

Tested on PCB versions V6.1 and V6.2


STM32/Temperature/Core/Inc/pcbversion.h
AndersJorgensenCA, line 1: Since you moved some pins, are the breaking major/minor still correct? (in main.h)

AndersJorgensenCA (approved):

STM32/Temperature/Core/Inc/pcbversion.h
TimotheDodinCA: Nope, it changes used pins depending on the PCB version

PR #260, opened 2026-05-29
Feature/ac efuse

LukeWalker-CA:
### Primary Changes

* Added an eFuse function that monitors an average of the current through a particular channel with a 1 second window. If the _average_ goes above a specified limit (pre-calibrated to 10, adjustable) then the channel will be disabled and an over-current flag set. The flag will stay active until the channel is enabled again.
  * Note: the _average_ behaviour means that with, e.g. a limit of 10 and an overcurrent limit of 50, the condition will be met after 0.2 secs. The intended behaviour is to allow small excursions beyond the fuse limit of 10 A temporarily (e.g. allow a 12 A load to be switched on 50% of the time)
* Unit tests for the above
* Specification document for AC board to help with AI assist coding

### Secondary Changes

* Formatting (review without final 2 commits to see non-formatting changes)
* Renamed ac unit test program to allow unit test debugging
* See https://github.com/copenhagenatomics/CA_Embedded_Libraries/pull/108

### Testing

- [x] unit tests
- [x] Towel rack


STM32/AC/specs/AC.md
```diff
@@ -0,0 +1,46 @@
...
+* Shall accept the following commands:
+  * Standard Copenhagen Atomics (CA) protocol commands
+  * Standard CA AC/DC protocol commands
+  * "fan on" - forces the fan to be on
+  * "fan off" - enables the fan to work in normal overheat control mode
+* Implementation of all on / all off:
+  * all on shall be specified with a duration (in seconds), otherwise return an undefined command message
+* Implementation of portState:
+  * portState "on" commands must specify a duration (in seconds), otherwise return an undefined command message
+  * If portState "on" commands specify a percentage, the AC signal must be PWM-d with the corresponding frequency (See below)
```
TimotheDodinCA: The percentage changes the duty cycle and not the frequency.

STM32/AC/specs/AC.md
```diff
@@ -0,0 +1,46 @@
...
+  * Any command to turn a pin on 100% (e.g. all on, pX on) shall take effect immediately
+  * Any command to turn a pin on 0% to 100% (e.g. pX on YY%) should take effect on the next PWM period
+  * Any command to turn a pin off, or on 0% (e.g. pX on 0%, all off) should take effect immediately
+* PWM
+  * AC PWM frequency should be 1 Hz, resolution 100 (to match zero crossings on a 50 Hz signal)
+    * Note: How to handle 60 Hz?
```
TimotheDodinCA, line 27: Hmm good question, might be worth noting that the grid frequency varies slighly during the day. So maybe 60 Hz will not be a problem. I'm unsure how this is implemented.
Today for example:

Image

STM32/AC/specs/AC.md
```diff
@@ -0,0 +1,46 @@
...
+* Overheat Control (if user has not forced fan on)
+  * If the heatsink temperature rises above 55 degC, enable the fan
+  * If the heatsink temperature rises above 70 degC, set the overheat error bit and enable the fan
+  * If the heatsink temperature sinks below 50 degC, disable the fan
+* E-fuse
+  * If the monitored current on a particular channel exceeds 10A average, measured over 1 PWM period, the channel shall be disabled and the overcurrent error bit set.
+  * The limit should be assessed and action taken as frequently as possible
+  * The overcurrent bit should be cleared at the next command affecting that channel
+  * The channel should continue to work as normal at the next command
+  * The current limit should be configurable, per channel, in calibration memory (calibration items 1, 2, 3 and 4)
```
TimotheDodinCA, line 39: Why should it be configurable? Isn't it made to protect the board?
If it is configurable, LoopControl will need to provide the user value at startup I guess

STM32/AC/Core/Src/flashHandler.c
```diff
@@ -43,7 +45,8 @@ typedef struct depositUnit_t {
 ** PUBLIC FUNCTION DEFINITIONS
 ***************************************************************************************************/
 
-faultInfo_t* fhGetFaultInfo() {return &(dpu.fault_info);}
+faultInfo_t* fhGetFaultInfo()    {return &(dpu.fault_info);}
+float*       fhGetCurrentLimits() {return dpu.currentLimits;}
```
TimotheDodinCA, line 49: Do we have any way of generating a fault, to verify that the flashHandler works as intended?

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -39,6 +40,9 @@
 
 #define USB_COMMS_TIMEOUT_MS     5000
 
+#define EFUSE_DEFAULT_CURRENT_LIMIT_A  10.0f
+#define EFUSE_MA_WINDOW                10   // samples per PWM period (10 Hz × 1 s)
```
TimotheDodinCA: The comment isn't that easy to understand (or mayeb it is me), as the grid frequency is 50 Hz, so the double is 100Hz.
I guess then your 10 Hz is 100 Hz / 10 ?
The good thing is that it will work as well with 60 Hz

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -109,6 +120,56 @@ static CAProtocolCtx caProto =
 ** PRIVATE FUNCTIONS
 ***************************************************************************************************/
 
+/*!
+** @brief Updates calibration memory with per-channel e-fuse current limits.
+**        Command format: "calibration p<N>,<limit_amps>,0,0"
```
TimotheDodinCA: Why isn't it the standard CA format CAL {channel},{limit},0 ?

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -109,6 +120,56 @@ static CAProtocolCtx caProto =
...
+*/
+static void ACcalibration(int noOfCalibrations, const CACalibration* calibrations) {
+    for (int i = 0; i < noOfCalibrations; i++) {
+        int port = calibrations[i].port;
+        if (port >= 1 && port <= AC_BOARD_NUM_PORTS && calibrations[i].alpha > 0) {
```
TimotheDodinCA, line 122: Maybe print an error message in case the conditions are not fullfilled

STM32/AC/Core/Src/ACBoard.c
```diff
@@ -334,6 +407,21 @@ static void CAallOn(bool isOn, int duration)
...
+/*!
+** @brief Clears the overcurrent fields for a specific port and clears the generic overcurrent 
+**        error if no ports are currently in overcurrent anymore.
+**
+** @param field The field to clear.
+*/
+static void clearOvercurrentFields(int field) {
+    bsClearField(AC_EFUSE_OVERCURRENT_Msk(field));
+
+    if(!(bsGetStatus() & AC_EFUSE_OVERCURRENT_ALL_Msk)) {
```
TimotheDodinCA: Doesn't bsGetField do that for you?

TimotheDodinCA (approved):
Nice implementation and unit tests!

STM32/AC/specs/AC.md
LukeWalker-CA: Poorly phrased. Reworded

STM32/AC/specs/AC.md
LukeWalker-CA: Yes - I think it will work (partly because 60 Hz is more frequent) but there might be some irritating phase beating.

STM32/AC/Core/Src/flashHandler.c
LukeWalker-CA: Not within the unit tests (yet?) but I have verified this on the benchtop. You can cast an out-of-memory address as a function pointer and then try to call it.

STM32/AC/Core/Src/ACBoard.c
LukeWalker-CA: 10 Hz * 1 s = 10 (unit-less. E.g. 10 samples). Let me know if you still think its unclear and I can try to think of a better re-wording

STM32/AC/Core/Src/ACBoard.c
LukeWalker-CA: It is! Stupid AI comment!

STM32/AC/specs/AC.md
LukeWalker-CA: Good point about LoopControl. I think for the minor FW effort, a configurable fuse limit might come in useful

PR #262, opened 2026-06-18
Feature/ac efuse

LukeWalker-CA:
### Primary Changes

* Updated current limit to 12 A (above fuse rating of 10 A, but the fuse should still be able to hold it indefinitely)
* Reduced trigger time to 0.1 s from 1 s
  * Removed moving average and all references to array-math

### Secondary Changes

* Formatting

### Testing

- [x] Unit tests
- [x] Towel Rack test


STM32/AC/Core/Src/ACBoard.c
```diff
@@ -40,8 +39,8 @@
 
 #define USB_COMMS_TIMEOUT_MS 5000
 
-#define EFUSE_DEFAULT_CURRENT_LIMIT_A 10.0f
-#define EFUSE_MA_WINDOW               10  // samples per PWM period (10 Hz × 1 s)
+/* According to the fuse document, 12 A can be held for > 10000 s */
+#define EFUSE_DEFAULT_CURRENT_LIMIT_A 12.0f
```
TimotheDodinCA: Maybe write RMS here as well

STM32/AC/specs/AC.md
```diff
@@ -32,7 +32,7 @@ This document contains the specification for the AC board/box project. Sub-compo
   * If the heatsink temperature rises above 70 degC, set the overheat error bit and enable the fan
   * If the heatsink temperature sinks below 50 degC, disable the fan
 * E-fuse
-  * If the monitored current on a particular channel exceeds 10A average, measured over 1 PWM period, the channel shall be disabled and the overcurrent error bit set.
+  * If the monitored current on a particular channel exceeds 12A average, measured over 0.1s, the channel shall be disabled and the overcurrent error bit set.
```
TimotheDodinCA: RMS and not average

STM32/AC/Core/Src/ACBoard.c
TimotheDodinCA: Could you use the CRC version of the flash functions here as well?

TimotheDodinCA (approved):
Looks good, just remember to replace avg with rms.
And maybe do the same CRC thing as DC

STM32/AC/Core/Src/ACBoard.c
LukeWalker-CA: Good idea, and implement uptime too

PR #263, opened 2026-06-24
DC Memory Map

LukeWalker-CA:
### Primary Changes

* Memory map fixes for multi-file programming

### Secondary Changes

* None

### Testing

- [x] Unit tests
- [x] Benchtop test

### DIFF
--- a/STM32/DC/Makefile
+++ b/STM32/DC/Makefile
@@ -197,7 +197,7 @@ LIBDIR =
 LDFLAGS = $(MCU) -specs=nano.specs -T$(LDSCRIPT) $(LIBDIR) $(LIBS) -Wl,-Map=$(BUILD_DIR)/$(TARGET).map,--cref -Wl,--gc-sections,--print-memory-usage -u_printf_float -u_scanf_float
 
 # default action: build all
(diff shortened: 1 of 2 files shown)


STM32/DC/Makefile
```diff
@@ -208,12 +209,18 @@ vpath %.c $(sort $(dir $(C_SOURCES)))
 # list of ASM program objects
 OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(ASM_SOURCES:.s=.o)))
 vpath %.s $(sort $(dir $(ASM_SOURCES)))
+OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(ASMM_SOURCES:.S=.o)))
```
TimotheDodinCA, line 212: Was that added automatically?

TimotheDodinCA (approved):

STM32/DC/Makefile
LukeWalker-CA: It was present in the GasSampler that you suggested I use as reference. I just tried auto-generating the makefile from scratch and ASMM was added automatically by STM32CubeMx but not ASMMC. It seems entirely plausible to me that it could have been added by STM32CubeMx automatically with a different software version / firmware package version.

I have tested it and it works, so I will merge.

PR #265, opened 2026-07-20
Patch/salt leak v1.3

TimotheDodinCA:
Patch release: should not be merged!

Changes:
- Using latest libraries
- StatusDef implementation
- OutputDef implementation


AndersJorgensenCA (approved):
Seems good to me

